Issue #18, opened 2017-04-04
Explain CubicFlatteningIter tolerance


dbrgn:
Hi. Great library, thanks a lot! :)

I used the Bézier part to do flattening of SVG paths. Could you maybe explain the tolerance value in the documentation in laymans terms? From my experiments, the lower the value the better the approximation. But what does the value actually mean? Does it have to do with the angle between two line segments?

(I should probably read the paper, but adding a short explanation to the docs would also help other people :))

nical:
Hi, thanks for the compliment! And thanks for filing the other issue, and sorry for not noticing it earlier.
The flattening tolerance is the maximum distance between the theoretical curve and the flattened approximation for an infinitely small line width. Ideally the this tolerance should take the curve width into account, but it does not, so if you want to be strictly correct you need to take it into account. I'll update the docs.

nical:
Fixed in #22.

PR #22, opened 2017-04-27
Improve the documentation of the lyon_bezier crate.

nical:
Addresses issue #18.


nical:
@dbrng This PR should address issue #18, let me know if the documentation could use some more clarification. The generated doc now looks like this http://imgur.com/a/pyM6G

dbrgn:
Awesome, that makes it *very* clear :) Thanks!

PR #26, opened 2017-05-15
Simplify the cubic bezier flattening code.

nical:
The cubic bezier flattening code finds up to two inflection points, and has special code to evaluate the approximation between the inflections and in the region very close to the inflections.

I think that it should be enough to only split the curve at the inflection points and evaluate the approximation on the sub curves without the special approximation of the region around the inflection point. This should result in an approximation that is almost as good (may generate an extra vertex but would not compromise the tolerance threshold) and greatly simplify the code.

This PR implements this simpler version of the flattening for both the iterator and callback versions. It would be best to figure out a way to verify that the approximation does not regress any important metric before removing the more complex version.

In the process of implementing this I also found and fixed a bug with the calculation of the second inflection point: once we have split the curve at the first inflection point, the position t2 of the second inflection point needs to be re-mapped into the space of the remaining portion of the curve.

### DIFF
--- a/bezier/src/flatten_cubic.rs
+++ b/bezier/src/flatten_cubic.rs
@@ -11,22 +11,6 @@ use super::{Point, CubicBezierSegment};
 use std::f32;
 use std::mem::swap;
 
-/// An iterator over a cubic bezier segment that yields line segments approximating the
-/// curve for a given approximation threshold.
-///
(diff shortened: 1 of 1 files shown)


nical:
This turned out to be a mistake, as explained in #110, fixed in #112.

Issue #48, opened 2017-05-27
Add a methods to compute the bounding rect of quadratic and cubic bezier curves


nical:
We first need to reexport euclid's rectangle type (`euclid::Rect<T>`) in bezier/src/lib.rs the same way we reexport the Point type, and add a method:

```rust
fn bounding_rect(&self) -> Rect
```

To `QuadraticBezierSegment` and `CubicBezierSegment`.

nical:
cc @o0Ignition0o

nical:
After a discussion with @o0Ignition0o, I like his suggestion to compute the minimum bounding rect, so I suggest having two methods:
```rust
fn bounding_rect(&self) -> Rect
```

which is a naive implementation that returns the minimum rectangle containing the control points of the curve, and

```rust
fn minimum_bounding_rect(&self) -> Rect
```

Which is slightly more expensive but returns a tighter rect. `find_y_inflection` should be useful for this, and for some reason `find_x_inflection` is missing but it could be added. `find_y_maximum` is also relevant, but `find_x_maximum`, `find_x_minimum` and `find_y_minimum` are missing too (should not be hard to add the missing methods taking inspiration from the existing ones).

o0Ignition0o:
This looks good, I'll give it a try ! I'm also considering rotating the curve to "flatten" the start and end points, before invoking find_y_maximum. I don't know if it would be the fastest implementation to get the minimum_bounding_rect, but I'll make a pr and see if I can improve it :)

PR #49, opened 2017-05-28
 #48 :  bounding_rect method for Quadratic and Cubic Bezier Segments

o0Ignition0o:
Added HasBoundingRect trait, and implemented the bounding_rect method for Quadratic and Cubic Bezier Segments.

This is the first part, I'll implement the other required methods for the issue #48 as soon as I find some time to do so :)

### DIFF
--- a/bezier/src/lib.rs
+++ b/bezier/src/lib.rs
@@ -70,6 +70,7 @@ mod cubic_to_quadratic;
 mod up_to_two;
 
 use std::mem::swap;
+use euclid::rect::rect;
 use flatten_cubic::{flatten_cubic_bezier, find_cubic_bezier_inflection_points};
 pub use flatten_cubic::CubicFlatteningIter;
 pub use cubic_to_quadratic::cubic_to_quadratic;
@@ -80,6 +81,12 @@ pub type Point = euclid::Point2D<f32>;
 
 /// Alias for ```euclid::Point2D<f32>```.
 pub type Vec2 = euclid::Point2D<f32>;
+
+/// Alias for ```euclid::Size2D<f32>```.
+pub type Size = euclid::Size2D<f32>;
+
+/// Alias for ```euclid::Rect<f32>```
+pub type Rect = euclid::Rect<f32>;
 
 /// A 2d curve segment defined by three points: the beginning of the segment, a control
 /// point and the end of the segment.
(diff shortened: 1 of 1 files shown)


bezier/src/lib.rs
```diff
@@ -81,6 +81,17 @@ pub type Point = euclid::Point2D<f32>;
...
 pub type Vec2 = euclid::Point2D<f32>;
 
+/// Alias for ```euclid::Size2D<f32>```.
+pub type Size = euclid::Size2D<f32>;
+
+/// Alias for ```euclid::Rect<f32>```
+pub type Rect = euclid::Rect<f32>;
+
+/// Allows to get a Rect structure which contains the Segment
+trait HasBoundingRect {
```
nical: The idea of adding a trait is interesting, but I'd prefer to do this separately. There are other common APIs between the different types of segments, and I'd like to brainstorm what the use cases are and how these traits should look like first (should we group them in some `Curve` trait? or should we have a trait for each group of common method? etc.).
Also, having a method exposed as trait implementations means you need to import the trait to use it which I find inconvenient, so I prefer the approach of implementing all methods on the type itself and then using them in trait impls.

So, could you remove the trait from this PR, I'll  open an issue so that we can brainstorm the API and we'll add traits as a followup.

o0Ignition0o: Thinking of it, I've seen lots of "common methods" among bezier curves, so I'll remove the trait, because this is a big commitment indeed. :)

bezier/src/lib.rs
```diff
@@ -249,6 +260,32 @@ impl QuadraticBezierSegment {
...
 }
 
+impl HasBoundingRect for QuadraticBezierSegment {
+    fn bounding_rect(&self) -> Rect {
+        let min_x = self.from.x.min(self.ctrl.x).min(self.to.x);
+        let max_x = self.from.x.max(self.ctrl.x).max(self.to.x);
+        let min_y = self.from.y.min(self.ctrl.y).min(self.to.y);
+        let max_y = self.from.y.max(self.ctrl.y).max(self.to.y);
+
+        return Rect::new(Point::new(min_x, min_y), Size::new(max_x - min_x, max_y - min_y));
```
o0Ignition0o: https://autumnai.github.io/cuticula/euclid/rect/struct.Rect.html I don't find any rect constructor which would allow me to do this, could you please point me to the right spot ?

nical: It looks like the euclid doc you are looking at is out of date, [the one in docs.rs](https://docs.rs/euclid/0.13.0/euclid/rect/fn.rect.html) has what we are looking for.

o0Ignition0o:
Thanks to your documentation, I've just pushed a modification using euclid::rect::rect instead. Is there an elegant way to import it ? If not I guess we should be fine :)

nical:
There isn't a more elegant way until euclid 0.13 gets published, until then that's fine.

bezier/src/lib.rs
```diff
@@ -247,6 +253,30 @@ impl QuadraticBezierSegment {
...
         return len;
     }
+
+    pub fn bounding_rect(&self) -> Rect {
+        let min_x = self.from.x.min(self.ctrl.x).min(self.to.x);
+        let max_x = self.from.x.max(self.ctrl.x).max(self.to.x);
+        let min_y = self.from.y.min(self.ctrl.y).min(self.to.y);
+        let max_y = self.from.y.max(self.ctrl.y).max(self.to.y);
+
+        return euclid::rect::rect(min_x, min_y, max_x - min_x, max_y - min_y);
```
nical: nit: just `use euclid::rect::rect` at the top of this file and then call `rect(...`.

o0Ignition0o:
This looks better, with one squashed commit :)

PR #55, opened 2017-05-28
Port UpToTwo to use ArrayVec.

emilio:
I was waiting for a LLVM build, so I decided to do this.

Sorry if this was a more easy issue more suitable for a newer rust community member, but I can't resist to negative diffs. Feel free to close it if that's the case :).


emilio:
Happy to bikeshed about stuff, in particular whether it looks cleaner to make `UpToTwo` something like:

```rust
struct UpToTwo<T>(ArrayVec<T>);
```

with auto-deref/deref-mut, so you can write:

```
return UpToTwo::two(a, b)
```

Which may or may not be cleaner. Or perhaps that may be another easy issue? *shrug*

nical:
Looks good to me. This is used in some pretty hot code so we may end up rolling back to a custom implem after the next round of profiling and optimization (UpToTwo wasn't particularly fast but the idea was to get to that later).
In the mean time this is both simpler and nicer so let's take it!

PR #56, opened 2017-05-28
#48 : maximum, minimum and inflection point methods for y and x for QuadraticBezierSegments.

o0Ignition0o:
Added maximum, minimum and inflection point methods for y and x for QuadraticBezierSegments. Wrote basic unit tests to reflect this.

I now need to implement the same methods for a cubic segment (the inflection point one will return up to two points) Before I define the minimum_bounding_rect one, and voila ! :)

Oh and by the way, I'm used to test proof any public function I write, and since the test suite is getting big, should we consider splitting the tests to the bottom of the .rs file, or event to a separate test_lib.rs file ?

### DIFF
--- a/bezier/src/lib.rs
+++ b/bezier/src/lib.rs
@@ -141,6 +141,19 @@ impl QuadraticBezierSegment {
         return if self.from.y > self.to.y { 0.0 } else { 1.0 };
     }
 
+    /// Find the advancement of the y-least position in the curve.
+    ///
+    /// This returns the advancement along the curve, not the actual y position.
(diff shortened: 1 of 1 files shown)


nical:
> Oh and by the way, I'm used to test proof any public function I write, and since the test suite is getting big, should we consider splitting the tests to the bottom of the .rs file, or event to a separate test_lib.rs file ?

If bezier/src/lib.rs becomes too big we can start by moving quadratic and cubic curves to their own files. I like to have tests close to the thing being tested, but if the amount of tests becomes so big that it is inconvenient (a very good problem to have!) we can reconsider.

PR #58, opened 2017-06-01
Bump euclid to 0.14

nical:
This is a very big change thanks to the separation of vector and point types in euclid 0.14. lyon used to sort of fake this by having separate `Point` and `Vec2` aliases to the same underlying Point2D type, but now this is properly implemented upstream, so no more confusing absolute (points) and relative (vectors) coordinates.


nical:
Ah, looks like another documentation commit got dragged in (I thought I had landed it already).

PR #60, opened 2017-06-11
Move the cubic and quadratic bezier code to their own files.

nical:
The bezier code will grow to incorporate more features as we add code to compute intersections, so let's split things up before it becomes impractical.
As we add intersection code, we'll need to move some most the triangle and line segment math code that lives in lyon_core to this crate, so I am thinking of turning this into a new crate (something like `lyon_geom`) with the math for 2d bezier curves, triangles, line segments, arcs and other geometric primitives.


o0Ignition0o:
Shall I wait for you to merge this pr before I continue working on the bounding_rect and bezier_curve_intersection issues ? :)

nical:
@o0Ignition0o If you want to submit a PR for the bounding rectangle parts that you already have I can rebase this one on top of it, otherwise I can land this now and let you rebase your stuff on top of this. As you prefer.

o0Ignition0o:
Please land this now, I'll rebase then, I won't be able to work on it tonight anyway :)

Issue #47, opened 2017-05-27
Implement 2d triangle vs triangle intersection test.


nical:
We need a function in core/src/math_utils.rs which would look like:

```rust
/// Returns true if the triangle t1 intersects t2.
pub fn triangle_intersection(t1: &[Point], t2: &[Point]) -> bool;
```

In case it helps, there is already a `triangle_contains` function to check if a triangle contains a point, and the function `segment_intersection` in math_utils.rs.

nical:
A triangle intersection algorithm is described here: https://pdfs.semanticscholar.org/ed24/eeb971d234594df5d1103283ae1a2bd200c8.pdf

nical:
Fixed in #63. That PR has a very simple and very slow implementation that is begging to be improved upon.

PR #63, opened 2017-06-13
Implement triangle-triangle intersection

nical:
A (very slow) initial implementation of the triangle-triangle intersection in the bézier crate.


nical:
Fixes #47.

PR #74, opened 2017-06-16
Cubic x y min  and max

o0Ignition0o:


o0Ignition0o:
Ok looks like my git jutsu skills are really poor indeed.

PR #80, opened 2017-06-23
Fix tests

nivkner:
Hi @nical, 

I noticed while running `cargo test --all` that the crate `lyon_svg` wasn't compiling because of a missing dependency on `lyon_extra`, and that tests were failing in `lyon`, so here's the fix. 

I assumed you don't need to test the ability to reexport the crates to I set the docs to `ignore`, however I don't really understand why they were failing. So if this is a different problem that needs to be addressed, let me know.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -140,14 +140,14 @@
 //!
 //! The meta-crate (`lyon`) mostly reexports the other lyon crates for convenience.
 //!
-//! ```
+//! ```ignore
 //! extern crate lyon;
 //! use lyon::tessellation::FillTessellator;
 //! ```
 //!
 //! Is equivalent to:
 //!
-//! ```
+//! ```ignore
 //! extern crate lyon_tessellation;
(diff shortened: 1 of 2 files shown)


svg/Cargo.toml
```diff
@@ -18,5 +18,6 @@ lyon_path_iterator = { version = "0.5.0", path = "../path_iterator" }
 lyon_path_builder = { version = "0.5.0", path = "../path_builder" }
 lyon_path = { version = "0.5.0", path = "../path" }
 lyon_core = { version = "0.5.0", path = "../core" }
+lyon_extra = { version = "0.5.0", path = "../extra" }
```
nical: Here I would prefer that you remove the `extern crate lyon_extra;` from svg/src/lib.rs rather than fix the dependency in cargo since the svg crate does not actually use lyon_extra.

nivkner: Done

PR #81, opened 2017-06-23
added normal calculation to fill_convex_polyline

kuxv:
@nical  Can you review this? I basically iterate over triplets of points so I always have data for normals. The first point being added is 2nd point of line, last is 1st point of line. Maybe I should add some unit test?

### DIFF
--- a/tessellation/src/basic_shapes.rs
+++ b/tessellation/src/basic_shapes.rs
@@ -716,38 +716,36 @@ pub fn stroke_circle<Output>(center: Point, radius: f32, tolerance: f32, output:
 
 /// Tessellate a convex polyline.
 // We insert 2nd point on line first in order to have the neighbours for normal calculation.
-pub fn fill_convex_polyline<'a, Iter, Output>(it: &'a Iter, output: &mut Output) -> Count
+pub fn fill_convex_polyline<Iter, Output>(mut it: Iter, output: &mut Output) -> Count
 where
-    &'a Iter: IntoIterator<Item = &'a Point>,
-    <&'a Iter as IntoIterator>::IntoIter: Clone,
+    Iter: Iterator<Item = Point> + Clone,
     Output: GeometryBuilder<FillVertex>,
 {
-    let mut it0 = it.into_iter();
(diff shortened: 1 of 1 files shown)


tessellation/src/basic_shapes.rs
```diff
@@ -715,39 +715,49 @@ pub fn stroke_circle<Output>(center: Point, radius: f32, tolerance: f32, output:
...
 /// Tessellate a convex polyline.
-///
-/// TODO: normals are not implemented yet.
-pub fn fill_convex_polyline<Iter, Output>(mut it: Iter, output: &mut Output) -> Count
+// We insert 2nd point on line first in order to have the neighbours for normal calculation.
+pub fn fill_convex_polyline<'a, Iter, Output>(it: &'a Iter, output: &mut Output) -> Count
 where
-    Iter: Iterator<Item = Point>,
+    &'a Iter: IntoIterator<Item = &'a Point>,
+    <&'a Iter as IntoIterator>::IntoIter: Clone,
```
nical: What requires the `Clone` trait here ?

nical: More generally, what motivated the change from taking an iterator to taking a &IntoIterator?

nical:
I prefer to keep using an `Iterator<Item = Point>` rather than an `&IntoIterator<Item = &Point>`. As far as I understand, `Iterator` provides with more flexibility for the caller (you can pass a chain of iterators as parameter), while `IntoIterator` has nicer syntactic sugar.
I am more inclined to keep the flexibility at the expense of the syntactic niceties.

You can implement this PR the same way with an `Iterator` that implements `Clone` and cloning it at the beginning instead of calling `into_iterator` thrice.

tessellation/src/basic_shapes.rs
kuxv: I went for IntoIter becuase I could not figure out how to pass the Iterator. But I figured it out finally ;)

kuxv:
I have reverted the original funtion signature and added the Clone trait requirement. I think I am getting better with Rust ;) However I do not think it is possible to get rid of the 2x clone() at the beginning.

PR #82, opened 2017-06-23
implement stroke_rounded_rectangle

### LINKED ISSUE #38: Implement rounded rectangle stroke tessellation.
The function `stroke_rounded_rectangle` in [basic_shapes.rs](https://github.com/nical/lyon/blob/master/tessellation/src/basic_shapes.rs) is currently unimplemented.

The other tessellation routines in the same file provide a good set of examples for how to tessellate specific shapes with the `GeometryBuilder` API.

nivkner:
Fixes #38

### DIFF
--- a/tessellation/src/basic_shapes.rs
+++ b/tessellation/src/basic_shapes.rs
@@ -588,11 +588,113 @@ fn stroke_border_radius<Output: GeometryBuilder<StrokeVertex>>(
 
 /// Tessellate the stroke of an axis-aligned rounded rectangle.
 pub fn stroke_rounded_rectangle<Output: GeometryBuilder<StrokeVertex>>(
-    _rect: &Rect,
-    _radii: &BorderRadii,
-    _output: &mut Output,
+    rect: &Rect,
+    radii: &BorderRadii,
+    tolerance: f32,
+    output: &mut Output,
 ) -> Count {
-    unimplemented!();
+    output.begin_geometry();
+
+    let w = rect.size.width;
+    let h = rect.size.height;
+    let x_min = rect.min_x();
+    let y_min = rect.min_y();
+    let x_max = rect.max_x();
+    let y_max = rect.max_y();
+    let min_wh = w.min(h);
+    let mut tl = radii.top_left.abs().min(min_wh);
+    let mut tr = radii.top_right.abs().min(min_wh);
+    let mut bl = radii.bottom_left.abs().min(min_wh);
+    let mut br = radii.bottom_right.abs().min(min_wh);
+
+    // clamp border radii if they don't fit in the rectangle.
+    if tl + tr > w {
+        let x = (tl + tr - w) * 0.5;
+        tl -= x;
+        tr -= x;
+    }
+    if bl + br > w {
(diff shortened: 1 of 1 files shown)


tessellation/src/basic_shapes.rs
```diff
@@ -588,11 +588,111 @@ fn stroke_border_radius<Output: GeometryBuilder<StrokeVertex>>(
...
+
+    let points = &[
+    [p1, p2],
+    [p5, p6],
+    [p3, p4],
+    [p0, p7],
+    ];
+
+    for side in points {
+        stroke_polyline((0..1).map(|p| {side[p]}), false, output);
```
nivkner: Couldn't turn the inner arrays into iterators no matter what I tried,
Always got
 ```
expected &TypedPoint2D
found TypedPoint2D
```
so I mapped to a range instead.
Just bringing this up because  I'd rather have this resolved if there is a simple solution to this that I just missed.

nivkner:
I see what mean, but I don't this case calls for a more manual approach. 

In the end all we're doing here is a simple "connect the dots"  operation on a continuous set of points, which to me sounds very much like an iterator. In fact, I was under the impression that iterators are the general-purpose way to produce shapes that don't fit any of the basic shapes. It just seems odd to me to turn this into a special case, just because it combines round and straight elements.

The real problem here, in my opinion, was actually `stroke_border_radius`, since I was too eager to call `stroke_polyline` I ended up fragmenting any shape using it that isn't a circle. So I propose instead to create a function to generate an iterator of the points along the path of the round border, which can be chained with others and only then call `stroke_polyline`. This could even be used be used to simplify the implementations of `fill_circle` and `fill_rounded_rectangle` using `fill_convex_polyline`.

That should guarantee that all of the parts connect, but of course, there is still the problem of the points on the round border not matching the other parts. For that the new function should do the following:

* It should depend on the points at it's edges for positioning and not on the angle. The reason for that is that it assumes the angle was calculated accurately to reflect the relative positions of the points, which can easily be false for borders whose angles are not known ahead of time (of course that is only relevant if the function is public, to help users create round shapes).

* It should not actually include the edges in the final iterator, that way when chaining with straight lines you won't get a duplicate point on the overall shape, which would be redundant even if the points perfectly overlap.

I think for now, given we are only working on shapes with known angles I'll just try to modify `stroke_border_radius` to fit the second criteria, but if your wondering: To fit the first criteria we could depict the edges as the points on a base of a Isosceles triangle whose side are the circle's radius and make the free-form parameter be the height. I thought of using the radius but that might make it unclear if the angle should be reflex. In this case we can use negative height.

nivkner:
Thanks for the review. I don't want to leave this pull request hanging while it it clear that the direction taken here isn't the way to go. I'll try your suggestion, but I also have other idea's I'd like to check. I'll send a new PR when I have something else to submit.

nical:
> I'll try your suggestion, but I also have other idea's I'd like to check.

Sounds good! I don't have any attachment to the suggestion I made so don't hesitate to go directly for another approach if it works well for you.

Sorry for the confusion, I am tired and I edited your comment instead of replying (I fixed it back).

PR #92, opened 2017-06-25
Added minimum_bounding_rect for quadratic bezier segments.

o0Ignition0o:


bezier/src/quadratic_bezier.rs
```diff
@@ -246,6 +247,16 @@ impl QuadraticBezierSegment {
 
         return rect(min_x, min_y, max_x - min_x, max_y - min_y);
     }
+
+    /// Returns the smallest rectangle the curve is contained in
+    pub fn minimum_bounding_rect(&self) -> Rect {
+        let min_x = self.from.x.min(self.sample(self.find_x_minimum()).x).min(self.to.x);
```
o0Ignition0o: Hmmm I think you're right, unless the curve is an actual flat line ? will change it if so :)

o0Ignition0o: Actually, the find_x_inflection() method returns 

> the x inflection point or None if this curve is x-monotone

So I guess I could write ` let min_x = self.from.x.min(self.sample(self.find_x_minimum()).x) ` instead of also comparing with "self.to.x". What d o you think ? :)

nical: (reviewing from my phone so I may be missing something) I think that `find_x/y_minimum/maximum` ready takes care of comparing the inflection point(s) against self.from/to (or should). 

Better than my initial suggestion : ```let min_x = self.sample_x(self.find_x_minimum);```

o0Ignition0o: Hey, You're correct, the finx_x_inflection() function does not return any advancement point if the curve is monotone, but the finx_x_maximum() function does the self.from.x and self.to.x check ! :)

PR #93, opened 2017-06-26
Implement round line joins

kaedroho:
This PR implements ``LineJoin::Round`` in the stroke tesselator (#5):

round-line-join

For tolerance to work properly, we need to know the line width in the tesselator. Would you be happy with me adding line width to the tesselator?

### DIFF
--- a/tessellation/src/path_stroke.rs
+++ b/tessellation/src/path_stroke.rs
@@ -337,11 +337,12 @@ impl<'l, Output: 'l + GeometryBuilder<Vertex>> StrokeBuilder<'l, Output> {
 
         self.length += (self.current - self.previous).length();
 
-        let (na, nb, maybe_nc) = get_angle_info(self.previous, self.current, to);
+        let normal = get_angle_normal(self.previous, self.current, to);
+
         let a_id = self.output.add_vertex(
             Vertex {
                 position: self.current,
-                normal: na,
+                normal: normal,
                 advancement: self.length,
                 side: Side::Left,
             }
@@ -349,27 +350,11 @@ impl<'l, Output: 'l + GeometryBuilder<Vertex>> StrokeBuilder<'l, Output> {
         let b_id = self.output.add_vertex(
             Vertex {
                 position: self.current,
-                normal: nb,
+                normal: -normal,
                 advancement: self.length,
                 side: Side::Right,
             }
         );
-
-        let (nc, c_id) = if let Some(n) = maybe_nc {
-            (
-                n,
-                self.output.add_vertex(
-                    Vertex {
-                        position: self.current,
-                        normal: n,
-                        advancement: self.length, // TODO
-                        side: Side::Left, // TODO
-                    }
-                )
-            )
-        } else {
-            (nb, b_id)
-        };
 
         if self.nth > 1 {
             self.output.add_triangle(self.previous_b_id, self.previous_a_id, b_id);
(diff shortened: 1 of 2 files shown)


nical:
Hi @kaedroho, thanks for this PR! At a glance it looks great, I am very jet-lagged so I will give it a more thorough look tomorrow.
It would be great if you could revert the changes to the gfx_advanced example (I really need to figure out a simple way so that we can fork some sort of playground without checking it into the repo and without having issues with the cargo workspace).

Also, you are right we need the expected line width to determine the join subdivision. I think that it's fine to add a `line_width` member to `StrokeOptions` which would default to 1.0 with a comment explaining that it is used to compute the subdivision. Or perhaps it would make things less confusing if the normals were scaled using this line_width parameter too (need to sleep on it).

kaedroho:
Thanks!

I'll add line_width to the tesselator but hold off on applying it to vertex positions until you've decided that's the approach you want to take. Personally, I think it would be a sensible default and we could always add a flag to switch it off for anyone who wants to apply the width in the shader.

tessellation/src/path_stroke.rs
```diff
@@ -437,21 +545,27 @@ fn get_angle_info(previous: Point, current: Point, next: Point) -> (Vec2, Vec2,
             }
         }
     };
-    return (inter - current, current - inter, None);
+    return inter - current;
+}
+
+/// Computes the max angle of a radius segment for a given tolerance
+pub fn compute_max_radius_segment_angle(radius: f64, tolerance: f64) -> f64 {
```
nical: nit: I think that f32 should be enough here.

tessellation/src/path_stroke.rs
```diff
@@ -337,81 +337,189 @@ impl<'l, Output: 'l + GeometryBuilder<Vertex>> StrokeBuilder<'l, Output> {
...
+                        advancement: self.length,
+                        side: front_side,
+                    }
+                );
+
+                (v, v)
+            }
+
+            LineJoin::Round => {
+                let max_radius_segment_angle = compute_max_radius_segment_angle(0.1 /* line width / 2 */, 0.001/*self.options.tolerance as f64*/);
```
nical: For the `line_width` stuff I'm thinking of going this way:
 - Have a `line_width: f32` member in `StrokeOptions` with default value `1.0` that we use for for the flattening.
 - Have an `apply_line_width: bool` member in `StrokeOptions`, with default value `true`. If true we apply the line width to positions, and if false we leave the position on the center of the line. Either way the normal stays the same.

I think that these two modes should give the most convenient thing to work with for both the simple use case and the one where we apply the line width in the shader. What do you think?

No need to implement `apply_line_width` in this PR, we can do this as a followup.

kaedroho: > I think that these two modes should give the most convenient thing to work with for both the simple use case and the one where we apply the line width in the shader. What do you think?

This sounds good to me

tessellation/src/path_stroke.rs
kaedroho: Changed in https://github.com/nical/lyon/pull/93/commits/06e7a2b8915b7f6210b9ff36adbd4f2d3450e2bc

kaedroho:
Thanks for the quick review!

I've just added ``line_width`` into ``StrokeOptions`` and plugged it in. This should be ready for merge now.

PR #95, opened 2017-06-27
Add apply_line_width to StrokeOptions

kaedroho:
This implements ``apply_line_width`` as per: https://github.com/nical/lyon/pull/93#discussion_r124292798

### DIFF
--- a/tessellation/src/path_stroke.rs
+++ b/tessellation/src/path_stroke.rs
@@ -116,6 +116,18 @@ impl StrokeTessellator {
     }
 }
 
+macro_rules! add_vertex {
+    ($builder: expr, $vertex: expr) => {{
+        let mut v = $vertex;
+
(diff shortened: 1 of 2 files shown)


tessellation/src/path_stroke.rs
```diff
@@ -116,6 +116,18 @@ impl StrokeTessellator {
     }
 }
 
+macro_rules! add_vertex {
+    ($builder: expr, $vertex: expr) => {{
+        let mut v = $vertex;
+
+        if $builder.options.apply_line_width {
+            v.position += v.normal * $builder.options.line_width;
```
nical: We should extrude the position by half of the line width on each side so that the total width of the curve is equal to `line_width`.

kaedroho:
Fixed, rebased and squashed. Thanks!

PR #97, opened 2017-06-28
[WIP] Cubic bezier min bounding rect

o0Ignition0o:
Went with an implementation so far but my unit test is not passing. I think I don't correctly understand what an inflection point is. Is my setup phase incorrect or is it the way I implement the method?

### DIFF
--- a/bezier/src/cubic_bezier.rs
+++ b/bezier/src/cubic_bezier.rs
@@ -183,6 +183,79 @@ impl CubicBezierSegment {
         find_cubic_bezier_inflection_points(self)
     }
 
+    /// Return local x extrema or None if this curve is monotone.
+    ///
+    /// This returns the advancements along the curve, not the actual x position.
+    pub fn find_local_x_extrema(&self) -> UpToTwo<f32> {
+        let mut ret = UpToTwo::new();
+        // See www.faculty.idc.ac.il/arik/quality/appendixa.html for an explanation
+        // The derivative of a cubic bezier curve is a curve representing a second degree polynomial function
+        // f(x) = a * x² + b * x + c such as :
+        let a = 3.0 * (self.to.x - 3.0 * self.ctrl2.x + 3.0 * self.ctrl1.x - self.from.x);
+        let b = 6.0 * (self.ctrl2.x - 2.0 * self.ctrl1.x + self.from.x);
+        let c = 3.0 * (self.ctrl1.x - self.from.x);
+
+        // If the derivative is a linear function
+        if a == 0.0 {
+            if b == 0.0 {
+                // If the derivative is a constant function
+                if c == 0.0 {
+                    ret.push(0.0);
+                }
+            } else {
+                ret.push(-c / b);
+            }
+            return ret;
+        }
+
+        fn in_range(t: f32) -> bool { t > 0.0 && t < 1.0 }
+
+        let discriminant = b * b - 4.0 * a * c;
+
+        // There is no Real solution for the equation
+        if discriminant < 0.0 {
+            return ret;
+        }
+
+        // There is one Real solution for the equation
+        if discriminant == 0.0 {
+            let t = -b / (2.0 * a);
+            if in_range(t) {
+                ret.push(t);
+            }
+            return ret;
+        }
+
+        // There are two Real solutions for the equation
+        let discriminant_sqrt = discriminant.sqrt();
+
+        let first_extremum = (-b - discriminant_sqrt) / (2.0 * a);
+        let second_extremum = (-b + discriminant_sqrt) / (2.0 * a);
+
+        if in_range(first_extremum) {
+            ret.push(first_extremum);
+        }
+
+        if in_range(second_extremum) {
+            ret.push(second_extremum);
+        }
+        ret
+    }
+
+    /// Return local y extrema or None if this curve is monotone.
+    ///
+    /// This returns the advancements along the curve, not the actual y position.
+    pub fn find_local_y_extrema(&self) -> UpToTwo<f32> {
+       let switched_segment = CubicBezierSegment {
+               from: yx(self.from),
+               ctrl1: yx(self.ctrl1),
+               ctrl2: yx(self.ctrl2),
+               to: yx(self.to),
+       };
+
+        switched_segment.find_local_x_extrema()
+    }
+
     /// Find the advancement of the y-most position in the curve.
     ///
     /// This returns the advancement along the curve, not the actual y position.
@@ -193,7 +266,7 @@ impl CubicBezierSegment {
             max_t = 1.0;
             max_y = self.to.y;
         }
-        for t in self.find_inflection_points() {
+        for t in self.find_local_y_extrema() {
             let point = self.sample(t);
             if point.y > max_y {
                 max_t = t;
@@ -213,7 +286,7 @@ impl CubicBezierSegment {
             min_t = 1.0;
             min_y = self.to.y;
         }
-        for t in self.find_inflection_points() {
+        for t in self.find_local_y_extrema() {
             let point = self.sample(t);
             if point.y < min_y {
                 min_t = t;
@@ -233,7 +306,7 @@ impl CubicBezierSegment {
             max_t = 1.0;
             max_x = self.to.x;
         }
-        for t in self.find_inflection_points() {
+        for t in self.find_local_x_extrema() {
             let point = self.sample(t);
             if point.x > max_x {
                 max_t = t;
@@ -243,9 +316,7 @@ impl CubicBezierSegment {
         return max_t;
     }
 
-    /// Find the advancement of the x-least position in the curve.
-    ///
-    /// This returns the advancement along the curve, not the actual x position.
+    /// Find the x-least position in the curve.
     pub fn find_x_minimum(&self) -> f32 {
         let mut min_t = 0.0;
         let mut min_x = self.from.x;
@@ -253,7 +324,7 @@ impl CubicBezierSegment {
             min_t = 1.0;
             min_x = self.to.x;
         }
-        for t in self.find_inflection_points() {
+        for t in self.find_local_x_extrema() {
             let point = self.sample(t);
             if point.x < min_x {
                 min_t = t;
@@ -263,11 +334,22 @@ impl CubicBezierSegment {
         return min_t;
     }
 
+    /// Returns a rectangle the curve is contained in
     pub fn bounding_rect(&self) -> Rect {
         let min_x = self.from.x.min(self.ctrl1.x).min(self.ctrl2.x).min(self.to.x);
         let max_x = self.from.x.max(self.ctrl1.x).max(self.ctrl2.x).max(self.to.x);
         let min_y = self.from.y.min(self.ctrl1.y).min(self.ctrl2.y).min(self.to.y);
         let max_y = self.from.y.max(self.ctrl1.y).max(self.ctrl2.y).max(self.to.y);
+
+        return rect(min_x, min_y, max_x - min_x, max_y - min_y);
+    }
+
+    /// Returns the smallest rectangle the curve is contained in
+    pub fn minimum_bounding_rect(&self) -> Rect {
+        let min_x = self.sample_x(self.find_x_minimum());
+        let max_x = self.sample_x(self.find_x_maximum());
+        let min_y = self.sample_y(self.find_y_minimum());
+        let max_y = self.sample_y(self.find_y_maximum());
 
         return rect(min_x, min_y, max_x - min_x, max_y - min_y);
     }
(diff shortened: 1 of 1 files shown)


nical:
Ooops! There's been some confusion about the inflection point indeed: Inflection points are points where the curve goes from turning in a direction to turning in another direction (my mental model of it is to imagine a bike riding along a curve and at some point the handlebars go from pointing to the left to pointing to the right).
X-inflection points are where the curve transitions from increasing along the x axis to decreasing and vis-versa.
What we are looking for in `find_x_minimum/maximum` is a local extremum of the curve along x, but the curves inflection points don't give us that (they give us points where the curvature is null), whereas x-inflection and y-inflection points would give us what we need.
So the `find_x/y_minimum/maximum` methods are not actually correct.

Good thing you are principled about unit testing and caught that!

nical:
To calculate the x-inflection points we need to solve the equation of the derivative `dx/dt = 0` with `x(t) = (1 - t)³ * from.x + 3 * (1 - t)² * t * ctrl1.x + 3 * t² * (1 - t) * ctrl2.x + t³ * to.x`.

o0Ignition0o:
Thanks a lot for your comment, I'll try to make the required changes asap ! Btw this probably means I'll have to do a followup on the Quadratic curves too isn't it ?

nical:
Cool, thanks a lot for looking into it!
I think that the quadratic bézier curve code is correct because x-/y-inflections points were computed using derivatives of x and y. Quadratic béziers don't have the other type of inflection point (if the curve starts turning to the left it always turns to the left and vis-versa, or in other words the sign of the curvature never changes) so there is less room for confusion.

o0Ignition0o:
Yeah indeed, I noticed my mistake as I just went through the code again, thanks for the hint ! :)

o0Ignition0o:
Ok I think I've found a way to solve this on paper, I'm gonna try to solve it in rust :)

o0Ignition0o:
Hey @nical , my find_y_extremums unit test is passing, but I don't understand why, mind giving me a hint on how I'm supposed to proceed plz ? thanks a lot ^^'

bezier/src/cubic_bezier.rs
```diff
@@ -183,6 +183,73 @@ impl CubicBezierSegment {
         find_cubic_bezier_inflection_points(self)
     }
 
+    /// Return local y extremum points or None if this curve is monotone.
+    ///
+    /// This returns the advancements along the curve, not the actual y position.
+    pub fn find_y_extremums(&self) -> UpToTwo<f32> {
```
nical: In order to have a coherent API with the quadratic bézier equivalent, this should be called `find_y_inflections`

We also need an equivalent method for x.

bezier/src/cubic_bezier.rs
```diff
@@ -243,26 +310,19 @@ impl CubicBezierSegment {
...
-    /// This returns the advancement along the curve, not the actual x position.
+    /// Find the x-least position in the curve.
     pub fn find_x_minimum(&self) -> f32 {
         let mut min_t = 0.0;
         let mut min_x = self.from.x;
         if self.to.x < min_x {
             min_t = 1.0;
             min_x = self.to.x;
         }
-        for t in self.find_inflection_points() {
```
nical: You should keep this loop, but replace `find_inflection_points` with `find_x_inflections`. Otherwise, if the control point is left of the start and end points (in which case the curve will go more to the left of from and to), this would not return the correct result.
`find_x_maximum` and `find_y_minimum/maximum` also need to be fixed up to use `find_x/y_inflections` instead of find_inflection_points`.

bezier/src/cubic_bezier.rs
```diff
@@ -272,6 +332,16 @@ impl CubicBezierSegment {
         return rect(min_x, min_y, max_x - min_x, max_y - min_y);
     }
 
+    /// Returns the smallest rectangle the curve is contained in
+    pub fn minimum_bounding_rect(&self) -> Rect {
+        let min_x = self.sample(self.find_x_minimum()).x;
+        let max_x = self.sample(self.find_x_maximum()).x;
+        let min_y = self.sample(self.find_y_minimum()).y;
+        let max_y = self.sample(self.find_y_maximum()).y;
```
nical: A little detail, but you can write `self.sample(self.find_x_minimum()).x` into `self.sample_x(self.find_x_minimum());` to avoid computing y (not extremely important since the code is already correct).

bezier/src/cubic_bezier.rs
```diff
@@ -106,12 +106,12 @@ impl CubicBezierSegment {
             ctrl2: ctrl1aa,
             to: ctrl1aaa,
         },
-        CubicBezierSegment {
-            from: ctrl1aaa,
-            ctrl1: ctrl2aa,
-            ctrl2: ctrl3a,
-            to: to,
-        });
+                CubicBezierSegment {
```
nical: nit: Not sure why this is getting an extra indentation level.

bezier/src/cubic_bezier.rs
```diff
@@ -183,6 +183,73 @@ impl CubicBezierSegment {
...
     }
 
+    /// Return local y extremum points or None if this curve is monotone.
+    ///
+    /// This returns the advancements along the curve, not the actual y position.
+    pub fn find_y_extremums(&self) -> UpToTwo<f32> {
+        let mut ret = UpToTwo::new();
+        // See www.faculty.idc.ac.il/arik/quality/appendixa.html for an explanation
+        // The derivative of a cubic bezier curve is a curve representing a second degree polynomial function
+        // f(x) = a * x² + b * x + c such as :
```
nical: To avoid confusion, let's say that it is the derivative of the curve projected on the y axis, and use `y` as the variable in the formula: `f(y) = a * y² + b * y + c` (because we also need the equivalent for the x axis).

bezier/src/cubic_bezier.rs
```diff
@@ -397,6 +483,25 @@ fn find_y_minimum_for_simple_cubic_segment() {
...
+fn find_y_extremums_for_simple_cubic_segment() {
+    let a = CubicBezierSegment {
+        from: Point::new(1.0, 1.0),
+        ctrl1: Point::new(1.0, 2.0),
+        ctrl2: Point::new(2.0, 1.0),
+        to: Point::new(2.0, 2.0),
+    };
+
+    let mut expected_y_extremums = UpToTwo::new();
+    expected_y_extremums.push(0.5);
```
nical: I don't think that 0.5 is an `y-inflection` point of the curve (although it is an inflection point). After drawing it on a piece of paper it looks like this curve is monotone (always increasing) along y.
I think that this curve should have no `y-inflections` (and also no `x-inflections`), even if, confusingly, 0.5 is a "normal" inflection point.

bezier/src/cubic_bezier.rs
```diff
@@ -183,6 +183,73 @@ impl CubicBezierSegment {
...
+                ret.push(t);
+            }
+            return ret;
+        }
+
+        // There are two Real solutions for the equation
+        let discriminant_sqrt = discriminant.sqrt();
+        println!("delta_sqrt : {} ", discriminant_sqrt);
+
+        let first_extremum = (-b - discriminant_sqrt) / 2.0 * a;
```
nical: It's easy to get confused about the precedence of operators. In order to avoid ambiguity, let's put some parentheses around `2.0 * a` here and below.

o0Ignition0o:
Thanks a lot for taking time to review my code, I'll edit it tomorrow :)

nical:
I am struggling to explain the distinction of x-inflection/y-inflection, and general inflection points of the curve without making it even more confusing, don't hesitate to ask for more details and/or help!

o0Ignition0o:
I'll give it a shot and let you know. Again thanks a lot for your patience ! :)

o0Ignition0o:
I fixed a couple of things, I guess I'm like halfway there. About the naming convention I'm acutally not looking for inflexion points (which are points where the Second derivative equals zero), but local extremums (which are points when whe derivative equals zero), so I'm not really sure we want to rename it, do we ?

nical:
> I'm acutally not looking for inflexion points (which are points where the Second derivative equals zero), but local extremums (which are points when whe derivative equals zero)

That's the distinction between inflection point of the parametric curve (second dervative), and x-inflection (where the curve x(t) or x(y) has an inflection which corresponds to first derivative).
But after all of the confusion around it, I'm starting to agree that calling it extremum instead of x/y-inflection is probably a good thing!

bezier/src/cubic_bezier.rs
```diff
@@ -183,6 +183,140 @@ impl CubicBezierSegment {
...
+                if c == 0.0 {
+                    ret.push(0.0);
+                }
+            } else {
+                ret.push(-c / b);
+            }
+            return ret;
+        }
+
+        fn in_range(t: f32) -> bool { t >= 0.0 && t < 1.0 }
```
nical: Let's use t strictly superior to zero here, since this is primarily useful to split the curve into monotone segments and computing the bounding rectangle of the curve, and in both situations there is no need to special-case of split the curve at t = 0.0.

o0Ignition0o:
Oh I think I understand now ! Well it's up to you, I believe both should have different names, but I'm not sure x/y_extremums is the most obvious name. Do you have any suggestions? Else I will rename it to x/y_inflection, and hope no confusion will be made :)

bezier/src/cubic_bezier.rs
```diff
@@ -183,6 +183,140 @@ impl CubicBezierSegment {
...
+        if in_range(second_extremum) {
+            ret.push(second_extremum);
+        }
+        ret
+    }
+
+    /// Return local y extremum points or None if this curve is monotone.
+    ///
+    /// This returns the advancements along the curve, not the actual y position.
+    pub fn find_y_extremums(&self) -> UpToTwo<f32> {
```
nical: In order to avoid duplicating this code, you could transpose the curve (flip x and y of all its points), and call find_x_extremums, [like solve_t_for_y does](https://github.com/nical/lyon/blob/9bc8ada976994d213667abf0d7a3f006b1f188b5/bezier/src/cubic_bezier.rs#L338).

bezier/src/cubic_bezier.rs
```diff
@@ -183,6 +183,140 @@ impl CubicBezierSegment {
         find_cubic_bezier_inflection_points(self)
     }
 
+    /// Return local x extremum points or None if this curve is monotone.
```
nical: nit: indentation looks like it's off by one level.

o0Ignition0o:
This seems fine, although I don't feel comfortable with the magic numbers I've just added to the minimum_bounding_rect_for_cubic_bezier_segment test, what do you think ?

`    let expected_minimum_bounding_rect = rect(0.0, -0.57735026, 2.0, 1.1547005); `

bezier/src/cubic_bezier.rs
```diff
@@ -369,6 +455,22 @@ fn bounding_rect_for_cubic_bezier_segment() {
...
 #[test]
+fn minimum_bounding_rect_for_cubic_bezier_segment() {
+    let a = CubicBezierSegment {
+        from: Point::new(0.0, 0.0),
+        ctrl1: Point::new(0.5, 2.0),
+        ctrl2: Point::new(1.5, -2.0),
+        to: Point::new(2.0, 0.0),
+    };
+    
+    let expected_minimum_bounding_rect = rect(0.0, -0.57735026, 2.0, 1.1547005);
```
nical: I would prefer to have a fuzzier test: something like checking that `rect(0.0, -0.6, 2.0, 1.2).contains_rect(&actual_minimum_bounding_rect)`. That way we should catch most regressions while not break the test every time we change the order of some operations and end up with slightly different rounding errors.

nical: Maybe also have a lower bound like also that `actual_minimum_bounding_rect.contains_rect(&rect(0.0, 0.5, 2.0, 1.0))` in addition to the other contains_rect test.

o0Ignition0o:
The test is fuzzed now ! :) plus I've squashed my commits :D Thanks a lot for your patience, I hope I'll be able to tackle harder features, feel free to ping me if you think an unassigned feature you have would match my skills and help me improve! 🙌

nical:
\o/

@o0Ignition0o If you are up for a bit more challenge (but reasonable still), I think that adding features to the stroke tessellator is fun to tackle because you get to actually render something and see the feature with your own eyes which is quite rewarding and the stroke tessellator is still reasonably understandable (while the fill tessellator is a really complex algorithm). For example #r4 and #35 would be fun to implement. As always, don't hesitate to ask for help!

o0Ignition0o:
Great, will try it then :D Oh and by the way the quadratic and cubic intersection features should be easy now that these methods are available !

PR #100, opened 2017-07-03
Compute the derivatives a cubic bézier curve.

nical:
The derivative is useful to solve t(x) for monotonic bézier curves using Newton's method among other things.

### DIFF
--- a/bezier/src/cubic_bezier.rs
+++ b/bezier/src/cubic_bezier.rs
@@ -1,4 +1,4 @@
-use {Point, Rect, rect, Transform2D};
+use {Point, Vec2, Rect, rect, Transform2D};
 use up_to_two::UpToTwo;
 use flatten_cubic::{flatten_cubic_bezier, find_cubic_bezier_inflection_points};
 pub use flatten_cubic::CubicFlatteningIter;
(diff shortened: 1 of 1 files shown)


nical:
cc @o0Ignition0o I had to find the derivative of the curve in `CubicBezierSegment::derivative_coefficients` which overlaps with what you are looking for in #97:

```
(- 3.0 * t2 + 6.0 * t - 3.0) * from
+ (9.0 * t2 - 12.0 * t + 3.0) * ctrl1
+ (- 9.0 * t2 + 6.0 * t) * ctrl2
+ (3.0 * t2) * to
```
Should give:

```
dx/dt = t * t  * (-3 * from.x + 9 * ctrl1.x - 9 * ctrl2.x - 3 * to.x)
      + t * (6 * from.x + 12 * ctrl1.x + 6 * ctrl2.x)
      + (-3 * from.x + 3 * ctrl1.x)
```

Which is equal to zero at the x-inflection points.

It wouldn't hurt to double-check, though.

o0Ignition0o:
hey, i came up with `  let a = 3.0 * (self.to.y - 3.0 * self.ctrl2.y + 3.0 * self.ctrl1.y - self.from.y);
        let b = 6.0 * (self.ctrl2.y - 2.0 * self.ctrl1.y + self.from.y);
        let c = 3.0 * (self.ctrl1.y - self.from.y);
 `

which does look like what you found, doesn't it ?

nical:
@o0Ignition0o great!

PR #111, opened 2017-07-07
Implement rounded rectangle stroke

### LINKED ISSUE #38: Implement rounded rectangle stroke tessellation.

nivkner:
Fixes #38.

The problem with the #82 was the segmentation of the shape, which could result in a mismatch between the different parts. This is addressed by producing one continuous shape, and by finding only the inner points in rounded borders (since the outer ones could be part of other paths).

### DIFF
--- a/tessellation/src/basic_shapes.rs
+++ b/tessellation/src/basic_shapes.rs
@@ -4,9 +4,10 @@
 
 use core::FlattenedEvent;
 use geometry_builder::{GeometryBuilder, Count, VertexId};
-use path_stroke::{StrokeOptions, StrokeTessellator};
+use path_stroke::{StrokeOptions, StrokeTessellator, StrokeBuilder};
 use math_utils::compute_normal;
 use math::*;
+use path_builder::BaseBuilder;
 use {FillVertex, StrokeVertex, Side};
 
 use std::f32::consts::PI;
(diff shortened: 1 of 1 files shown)


tessellation/src/basic_shapes.rs
```diff
@@ -610,11 +611,124 @@ fn stroke_border_radius<Output: GeometryBuilder<StrokeVertex>>(
...
+    // left
+    let p0 = point(x_min, y_min + tl);
+    let p7 = point(x_min, y_max - bl);
+
+    // right
+    let p3 = point(x_max, y_min + tr);
+    let p4 = point(x_max, y_max - br);
+
+    let sides = &[
+    [p1, p2],
```
nical: nit: let's indent this one extra level.

tessellation/src/basic_shapes.rs
```diff
@@ -610,11 +611,124 @@ fn stroke_border_radius<Output: GeometryBuilder<StrokeVertex>>(
...
+        builder.line_to(p2);
+        stroke_border_radius_build(
+            centers[0],
+            angles[0],
+            radii[0],
+            nums.next().unwrap(),
+            &mut builder,
+        );
+        for i in 1..4 {
+            builder.line_to(sides[i][0]);
```
nical: Isn't the end of the previous iteration already going to finish at this position? If we can make it so, then you could even move the `line_to` and the `stroke_border_radius_build` above into the loop (and just have the `move_to` outside the loop).

nivkner:
moved the first `stroke_border_radius_build` and `line_to` into the loop as requested.

nical:
Thanks a lot!

PR #113, opened 2017-07-07
stroke_border_radius fixes

nivkner:
hi @nical,

There are  a few things I'd like to fix now that #111 got merged, namely:

* Removing  `stroke_border_radius`. I was too eager to call `stroke_polyline` on this one that it ended up being a very special case, so I'm replacing it with `stroke_border_radius_build` which I'll rename to it's shorter version now that they don't clash.
* I didn't realize before that the use of `log2` was because of the recursion and so every time one would use these functions, they would get a lot less segments than intended.
* Re-implement `stroke_circle` with the new `stroke_border_radius`

### DIFF
--- a/tessellation/src/basic_shapes.rs
+++ b/tessellation/src/basic_shapes.rs
@@ -583,32 +583,6 @@ fn fill_border_radius<Output: GeometryBuilder<FillVertex>>(
     );
 }
 
-// tessellate the stroke of rounded corners.
-fn stroke_border_radius<Output: GeometryBuilder<StrokeVertex>>(
-    center: Point,
-    angle: (f32, f32),
-    radius: f32,
-    resolution: u32,
-    output: &mut Output
-) {
-    if resolution == 0 {
-        return;
-    }
-
-    let angle_size = (angle.0 - angle.1).abs();
-    let starting_angle = angle.0.min(angle.1);
-    let closed = angle.0 % (2.0 * PI) == angle.1 % (2.0 * PI);
-
-    let points = (0..resolution).map(|i| {
-        let new_angle = i as f32 * (angle_size) / resolution as f32 + starting_angle;
-        let normal =
-        vec2(new_angle.cos(),
-        new_angle.sin());
-        center + normal * radius
-    });
-    stroke_polyline(points, closed, output);
-}
-
 /// Tessellate the stroke of an axis-aligned rounded rectangle.
 pub fn stroke_rounded_rectangle<Output: GeometryBuilder<StrokeVertex>>(
     rect: &Rect,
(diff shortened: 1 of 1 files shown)


tessellation/src/basic_shapes.rs
```diff
@@ -803,26 +774,31 @@ pub fn stroke_circle<Output>(center: Point, radius: f32, tolerance: f32, output:
     }
 
     let angle = (0.0, 2.0 * PI);
+    let starting_point = center + vec2(1.0, 0.0) * radius;
 
     let arc_len = 0.5 * PI * radius;
```
nical: The error was already there, but I think that this should be `2.0 * PI * radius` since this covers the entire circle and not just a corner.

nivkner: ok, changed that for every instance of  `arc_len`, since they are all based on the same procedure.

tessellation/src/basic_shapes.rs
```diff
@@ -515,7 +515,7 @@ pub fn fill_rounded_rectangle<Output: GeometryBuilder<FillVertex>>(
         let radius = radii[i];
         if radius > 0.0 {
 
-            let arc_len = 0.5 * PI * radius;
+            let arc_len = 2.0 * PI * radius;
```
nical: I think that this was correct here as `0.5 * PI * radius`, since we are calling this 4 times (one for each quarter circle).

tessellation/src/basic_shapes.rs
```diff
@@ -692,12 +666,9 @@ pub fn stroke_rounded_rectangle<Output: GeometryBuilder<StrokeVertex>>(
 
     let mut nums = radii.iter().map(|&radius| {
         if radius > 0.0 {
-
-            let arc_len = 0.5 * PI * radius;
+            let arc_len = 2.0 * PI * radius;
```
nical: Same here, `0.5 * PI * radius` should be correct because each corner is a quarter of a circle.

tessellation/src/basic_shapes.rs
```diff
@@ -771,7 +742,7 @@ pub fn fill_circle<Output: GeometryBuilder<FillVertex>>(
         (PI * 0.5, PI),
     ];
 
-    let arc_len = 0.5 * PI * radius;
+    let arc_len = 2.0 * PI * radius;
```
nical: Same there: the arc length is for a quarter of the circle and we do each quarter separately, so `0.5 * PI * radius` should be good.

tessellation/src/basic_shapes.rs
```diff
@@ -803,26 +774,31 @@ pub fn stroke_circle<Output>(center: Point, radius: f32, tolerance: f32, output:
     }
 
     let angle = (0.0, 2.0 * PI);
+    let starting_point = center + vec2(1.0, 0.0) * radius;
 
-    let arc_len = 0.5 * PI * radius;
+    let arc_len = 2.0 * PI * radius;
```
nical: This one, however, is indeed meant to use `2.0 * PI * radius` since the `stroke_border_radius` call below handles the entire circle so we have to compute the number of segments for the entire circle rather than a quarter of it like we do in the other places.

nivkner:
fixed, though this really seems to call for a more general approach, given how interdependent many of these values are. What would be the the performance cost of deriving the `arc_len` from the angle for example?

nical:
Or wrap part of the computation in a function that's explicit enough and explicitly device whn we deal with quarters, like `let arc_len = circle_length(radius) / 4.0; // for the "quarter" cases` which would make it more obvious I suppose. I am not too found generalizing in this particular case because we'd make the math more complicated than it needs to be (we really only need full circle and quarter of a circle for all of the use cases of this file).

Issue #88, opened 2017-06-24
API guidelines: return error type from functions instead of empty struct.


nical:
This is a sub-issue for #44.

We should audit the docs/code of each lyon crate and check that [error types are meaningful`](https://github.com/brson/rust-api-guidelines#c-meaningful-err).

I know that there are a some shortcomings in this respect in the lyon crates. For example in a lot of places the error type is an empty struct and the reasons of errors are not properly communicated.

nical:
Fixed by @Noxivs in #88.

PR #139, opened 2017-08-10
Add meaningful errors for the svg crate #88

Noxivs:
First little contribution, I will mainly focus on API guidelines issues for now.

This is a pull request for #88 but there could be additional changes.

If there is no point to group parser errors in one enum, I can split them.

I also checked other lyon crates and I didn't find similar cases. If you find more, let me know.

### DIFF
--- a/svg/src/parser/error.rs
+++ b/svg/src/parser/error.rs
@@ -1,0 +1,24 @@
+
+use std::fmt;
+use svgparser::Error;
+
+/// Errors which can occur when attempting to parse a token.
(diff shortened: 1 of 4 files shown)


nical:
Welcome to the [list of contributors](https://github.com/nical/lyon/wiki/Contributors)!

> I also checked other lyon crates and I didn't find similar cases. If you find more, let me know.

Looks like there isn't any other outstanding `Result<Foo, ()>`.

Issue #87, opened 2017-06-24
API guidelines: iterator type names should match the methods that produce them.


nical:
This is a sub-issue for #44.

We should audit the docs/code of each lyon crate and check that [iterator type names match the methods that produce them`](https://github.com/brson/rust-api-guidelines#c-iter-ty).

I know that most of the iterators in lyon crates don't follow this convention.

nical:
Here is a list of the methods/iterators and the names they should change to:

In `lyon_bezier` (@anna-liao urrently working on it):

 - [ ] `CubicBezierSegment::flattening_iter`/`CubicFlatteningIter` -> `CubicBezierSegment::flattened`/`cubic_bezier::Flattened`
 - [ ] `QuadraticBezierSegment::flattening_iter`/`QudraticFlatteningIter` -> `QuadraticBezierSegment::flattened`/`quadratic_bezier::Flattened`

In `lyon_path`:
 - [ ] `PathIter` -> `Iter`

 
In `lyon_path_iterator`:

- [ ] `FlattenginIter` -> `Flattened` (the method name is already `flattened`)
- [ ] `PathStateSvgIter` -> ??? (unclear, ideally `PathIter` but there is a collision with `PathStateIter` which should also be `PathIter`. Maybe splitting this file into sub modules would help.
- [ ] `PathStateIter` -> ??? (same as above)
- [ ] `SvgToPathIter` -> ??? (same as above)

PR #144, opened 2017-08-16
[WIP] API guidelines: iterator type names should match the methods that produce them

anna-liao:
working on https://github.com/nical/lyon/issues/87.

I changed the iterators to match the method names.  However, I run into this error when trying to use Flattened for both classes.  Sorry, I'm new to Rust, so perhaps this issue was too challenging for me.

  cargo test --all
   Compiling lyon_bezier v0.7.1 (file:///Users/annaliao/repos/lyon/bezier)
error[E0252]: a type named `Flattened` has already been imported in this module
   --> bezier/src/lib.rs:103:44
    |
102 | pub use quadratic_bezier::{QuadraticBezierSegment, Flattened};
    |                                                    --------- previous import of `Flattened` here
103 | pub use cubic_bezier::{CubicBezierSegment, Flattened};
    |                                            ^^^^^^^^^ `Flattened` already imported

error[E0252]: a type named `Flattened` has already been imported in this module
   --> bezier/src/lib.rs:103:44
    |
102 | pub use quadratic_bezier::{QuadraticBezierSegment, Flattened};
    |                                                    --------- previous import of `Flattened` here
103 | pub use cubic_bezier::{CubicBezierSegment, Flattened};
    |                                            ^^^^^^^^^ `Flattened` already imported

error: aborting due to previous error

error: Could not compile `lyon_bezier`.
Build failed, waiting for other jobs to finish...
error: aborting due to previous error

error: build failed

### DIFF
--- a/bezier/src/cubic_bezier.rs
+++ b/bezier/src/cubic_bezier.rs
@@ -2,7 +2,7 @@ use {Point, Vec2, Rect, rect, Line, LineSegment, Transform2D};
 use up_to_two::UpToTwo;
 use arrayvec::ArrayVec;
 use flatten_cubic::{flatten_cubic_bezier, find_cubic_bezier_inflection_points};
-pub use flatten_cubic::CubicFlatteningIter;
+pub use flatten_cubic::Flattened;
 pub use cubic_to_quadratic::cubic_to_quadratic;
 use monotone::{XMonotoneParametricCurve, solve_t_for_x};
 use utils::cubic_polynomial_roots;
@@ -161,8 +161,8 @@ impl CubicBezierSegment {
 
     /// Returns the flattened representation of the curve as an iterator, starting *after* the
     /// current point.
-    pub fn flattening_iter(&self, tolerance: f32) -> CubicFlatteningIter {
-        CubicFlatteningIter::new(*self, tolerance)
+    pub fn flattened(&self, tolerance: f32) -> Flattened {
+        Flattened::new(*self, tolerance)
     }
 
     /// Iterates through the curve invoking a callback at each point.
--- a/bezier/src/flatten_cubic.rs
+++ b/bezier/src/flatten_cubic.rs
@@ -17,7 +17,7 @@ use std::mem::swap;
 ///
 /// The iterator starts at the first point *after* the origin of the curve and ends at the
 /// destination.
-pub struct CubicFlatteningIter {
+pub struct Flattened {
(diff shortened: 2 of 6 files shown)


nical:
@anna-liao don't worry, lets walk through this together. Rust's module/use system is a bit confusing when discovering the language.

This error is happening because both the cubic_bezier and quadratic_bezier's `Flattened` types are imported in the root namespace of lyon_bezier through the ```pub use cubic_bezier::{CubicBezierSegment, Flattened};``` and ```pub use quadratic_bezier::{QuadraticBezierSegment, Flattened};``` use statements in lib.rs. But we are not allowed to have two things with the same name imported in a module.

For example:

```rust
// in a.rs:
pub struct Foo {}

// in b.rs:
pub struct Foo {}

// in c.rs
use a::Foo;
use b::Foo; // Error! can't have two things named Foo at the same place
...
let f = Foo; // can't know which Foo we are referring to
```

The above does not work, but it can be fixed by doing this:

```rust
// in a.rs:
pub struct Foo {}

// in b.rs:
pub struct Foo {}

// in c.rs
use a;
use b;
...
// a::Foo and b::Foo are different "names" so it works.
let f1 = a::Foo;
let f2 = b::Foo;
```

`quadratic_bezier` and `cubic_bezier` are equivalent to `a` and `b` in the example above. To fix our problem, we have to:
 -  turn these into ```pub use quadratic_bezier::QuadraticBezierSegment;``` and ```pub use cubic_bezier::CubicBezierSegment;``` to stop importing the `Flattened` name twice into lyon_bezier,
 - make the `mod cubic_bezier;` and `mod quadratic_bezier;` public by turning them into `pub mod cubic_bezier;` and `pub mod quadratic_bezier;`,
 - then, in code that uses them (for example in path_iterator.rs), refer to `quadratic_bezier::Flattened` instead of `QuadraticFlatteningIter` (and similar thing for the cubic one).

I hope this isn't too confusing. Don't hesitate to ask for more help.

anna-liao:
@nical Thanks for the explanation!  I made the changes listed for https://github.com/nical/lyon/issues/87, except `PathStateSvgIter` and `SvgToPathIter` since I'm not sure how to split into submodules.  Just let me know if I should squash the commits also.

nical:
Thanks again for the PR and welcome to the [list of contributors](https://github.com/nical/lyon/wiki/Contributors). Hope to hear from you again!

PR #154, opened 2017-08-27
Updated the path_stroke.rs documentation.

o0Ignition0o:
The dependencies used in the documentation were a bit rusty (please laugh), so I struggled a bit to get it working.
I guess this could be useful :)


o0Ignition0o:
Hmm it turns out the old one did pass, nevermind x)

Issue #34, opened 2017-05-21
Implement bevel line joins.


nical:
The path stroke tessellator has a very naive implementation of line joins that does not support bevel joins.

The relevant par of the SVG specification is here: https://svgwg.org/specs/strokes/#LineJoin

This should be implemented in [tessellation/src/path_stroke.rs](https://github.com/nical/lyon/blob/master/tessellation/src/path_stroke.rs).

See the (work in progress) [stroke tessellation wiki page](https://github.com/nical/lyon/wiki/Stroke-tessellation) for information about how the tessellator works.

o0Ignition0o:
Ok I want to give this one a shot when I have time, can you please assign me ? :)

o0Ignition0o:
Hey @nical , I've just found a stackexchange post https://graphicdesign.stackexchange.com/questions/59018/what-is-a-bevel-join-of-two-lines-exactly-illustrator which claims : 

> The two points are orthogonal to their respective segments and half the thickness away from them

Can you please confirm the is the approach I'm supposed to take in order to tackle the issue ? 

https://i.stack.imgur.com/TWRGI.png 

Thanks a lot :)

nical:
Hi @o0Ignition0o, yes, this looks right. The other join on this stackexchange post is a miter-clip join which we should also get to eventually, but first things first :)

EloD10:
I just see that : "o0Ignition0o commented 9 days ago ", I thought I was reading 9 months, i'm so sorry if you were working on that issue @o0Ignition0o

o0Ignition0o:
It's ok ! nice one btw, I was trying hard to find a way to cover the  class with unit tests, so it should be fine :)

PR #158, opened 2017-09-04
Implement bevel line joins

EloD10:
#34 
*cross fingers*

### DIFF
--- a/bench/path_fill/src/fill_bench.rs
+++ b/bench/path_fill/src/fill_bench.rs
@@ -32,8 +32,7 @@ fn logo_tess_no_intersection(bench: &mut Bencher) {
     let path = path.build();
 
     let mut tess = FillTessellator::new();
-    tess._handle_intersections = false;
-    let options = FillOptions::default();
+    let options = FillOptions::default().assume_no_intersections();
     let events = FillEvents::from_iterator(path.path_iter().flattened(0.05));
 
     bench.iter(|| {
--- a/tessellation/src/basic_shapes.rs
+++ b/tessellation/src/basic_shapes.rs
@@ -715,7 +715,7 @@ pub fn fill_ellipse<Output: GeometryBuilder<FillVertex>>(
 
     return FillTessellator::new().tessellate_events(
         &events,
-        &FillOptions::tolerance(tolerance),
+        &FillOptions::tolerance(tolerance).assume_no_intersections(),
         output,
     ).unwrap();
 }
--- a/tessellation/src/path_fill.rs
+++ b/tessellation/src/path_fill.rs
@@ -183,7 +183,7 @@ pub struct FillTessellator {
     previous_position: TessPoint,
     error: Option<FillError>,
     log: bool,
-    pub _handle_intersections: bool,
+    handle_intersections: bool,
 }
 
 impl FillTessellator {
@@ -198,7 +198,7 @@ impl FillTessellator {
             previous_position: TessPoint::new(FixedPoint32::min_val(), FixedPoint32::min_val()),
             error: None,
             log: false,
-            _handle_intersections: true,
+            handle_intersections: true,
         }
     }
 
@@ -253,6 +253,8 @@ impl FillTessellator {
         if options.fill_rule != FillRule::EvenOdd {
             println!("warning: Fill rule {:?} is not supported yet.", options.fill_rule);
         }
+
+        self.handle_intersections = !options.assume_no_intersections;
 
         self.begin_tessellation(output);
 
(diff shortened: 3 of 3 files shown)


nical:
I gave this a quick try and the resulting tessellation looks incorrect.

You can try it out by modifying the `gfx_advanced` example and replacing the code that builds the logo towards the beginning of the `main` function with:

```rust
    let mut builder = SvgPathBuilder::new(Path::builder());
    builder.move_to(point(0.0, 0.0));
    builder.line_to(point(100.0, 0.0));
    builder.line_to(point(100.0, 100.0));
    builder.close();

    let path = builder.build();
```

And further down where the stroke tessellator is used, replace:

```rust
    StrokeTessellator::new().tessellate_path(
        path.path_iter(),
        &StrokeOptions::tolerance(0.022).dont_apply_line_width(),
        &mut BuffersBuilder::new(&mut cpu.strokes, WithId(logo_stroke_id.element)),
    );
```

with:

```rust
    use lyon::tessellation::LineJoin;
    StrokeTessellator::new().tessellate_path(
        path.path_iter(),
        &StrokeOptions::tolerance(0.022).dont_apply_line_width().with_line_join(LineJoin::Round),
        &mut BuffersBuilder::new(&mut cpu.strokes, WithId(logo_stroke_id.element)),
    );
```

Then `cargo run` from the `gfx_advanced` directory. This will render a bunch of filled triangle shapes some of which are moving. The one that's not moving has a stroke and the bevel does not look right at the moment.

I also tried the same shape with the round join and it appears to have some issues as well. I'll look into fixing that as soon as I have a bit of time, until then best to take the round join implementation with a grain of salt.

nical:
@EloD10 I figured out the problems with round joins and landed fixes in #159. If you rebase your patch on top of the master branch some of the issues (that were not related to your PR) should be gone.

tessellation/src/path_stroke.rs
```diff
@@ -660,6 +660,44 @@ impl<'l, Output: 'l + GeometryBuilder<Vertex>> StrokeBuilder<'l, Output> {
...
+                let mut last_vertex = add_vertex!(
+                    self,
+                    Vertex {
+                        position: self.current,
+                        normal: normal,
+                        advancement: self.length,
+                        side: front_side,
+                        }
+                );
+                let back_vertex = last_vertex;
```
nical: I think that `back_vertex` here should be the one already created, instead of using `last_vertex` (otherwise it means `last_vertex` and `back_vertex` are the same vertex and the resulting triangle is infinitely thin).

tessellation/src/path_stroke.rs
```diff
@@ -660,6 +660,44 @@ impl<'l, Output: 'l + GeometryBuilder<Vertex>> StrokeBuilder<'l, Output> {
...
+                        advancement: self.length,
+                        side: front_side,
+                        }
+                );
+
+                // Compute 2 normal on the right and left side  
+                let normal = if front_side == Side::Right {
+                    vec2(a_line.y, -a_line.x).normalize() * 0.5
+                } else {
+                    vec2(-a_line.y, a_line.x).normalize() * 0.5
```
nical: Part of the fix of from #159 was to remove these extra `* 0.5` (which are already applied in the add_vertex macro where needed), so you can remove it here and above.
Also, in that PR `a_line` got renamed into `prev_tangent` and `b_line` into `next_tangent`, and I changed them to be already normalized so you won't need to do it here after rebasing.

tessellation/src/path_stroke.rs
```diff
@@ -660,6 +660,44 @@ impl<'l, Output: 'l + GeometryBuilder<Vertex>> StrokeBuilder<'l, Output> {
...
+                        normal: normal,
+                        advancement: self.length,
+                        side: front_side,
+                        }
+                );
+                let back_vertex = last_vertex;
+                
+                self.output.add_triangle(back_vertex, current_vertex, last_vertex);
+
+                (back_vertex, last_vertex)
```
nical: Another thing that changed in #159 is that `self.prev_normal` has to be updated to the normal of the last or back vertex (depending on which side they are) by the code that handles the join, so you'll have to do something similar (see the new round join code).

tessellation/src/path_stroke.rs
```diff
@@ -660,6 +660,44 @@ impl<'l, Output: 'l + GeometryBuilder<Vertex>> StrokeBuilder<'l, Output> {
...
             }
 
+            LineJoin::Bevel => {
+            
+                // First we take our initial vertex for compute our triangle
+                let current_vertex = add_vertex!(
+                    self,
+                    Vertex {
+                        position: self.current,
+                        normal: normal,
```
nical: I think that the normal of this vertex might not be equal to `normal` (which is the average normal of the join), but rather a vector perpendicular to `prev_tangent`.

Issue #165, opened 2017-09-08
Broken links in lyon_tessellation doc


nical:
See: https://docs.rs/lyon_tessellation/0.7.4/lyon_tessellation/

In the examples section, the links "tessellating path fills" and "tessellating path strokes" are broken.

PR #166, opened 2017-09-09
Implement bevel line joins

EloD10:
#34 
As you see, all works :
yeah

but, when i press "z", we can see that : 
noooo

It's the same bug that Round join have. I will take some times for see what doesen't work. It's something related to sides i think.

### DIFF
--- a/tessellation/src/lib.rs
+++ b/tessellation/src/lib.rs
@@ -168,8 +168,8 @@
 //!
 //! ## Examples
 //!
-//! - [tessellating path fills](path_fill/index.html#examples).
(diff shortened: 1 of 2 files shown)


nical:
Looks good! the reason the rendering looks off when you press 'z' a lot is that the line width becomes negative and the shader starts pushing vertices in the opposite direction. This is not related to your bevel implementation, it's really just the demo which should prevent the width from being negative since it makes no sense.

Could you squash the three commits into one commit? After that we are good to merge.

EloD10:
Done! Also it will resolve this issue #165 normally.

nical:
Thanks!

PR #192, opened 2017-10-21
#35 Implement miter_limit_join

o0Ignition0o:
This is Working on a function signature, before miter_limit_join_implementation (and tests).
I guess I should remove the magic number 4, to use an enum or something instead.

This should handle https://github.com/nical/lyon/issues/35 some day x)

### DIFF
--- a/tessellation/src/path_stroke.rs
+++ b/tessellation/src/path_stroke.rs
@@ -1,6 +1,6 @@
 use math::*;
 use core::FlattenedEvent;
-use bezier::{QuadraticBezierSegment, CubicBezierSegment};
+use bezier::{QuadraticBezierSegment, CubicBezierSegment, LineSegment};
 use bezier::utils::{normalized_tangent, directed_angle, fast_atan2};
 use geometry_builder::{VertexId, GeometryBuilder, Count};
 use basic_shapes::circle_flattening_step;
@@ -587,13 +587,12 @@ impl<'l, Output: 'l + GeometryBuilder<Vertex>> StrokeBuilder<'l, Output> {
             }
         );
 
-        let limit = self.options.miter_limit;
         if prev_tangent.dot(next_tangent) >= 0.95 {
             // The two edges are almost aligned, just use a simple miter join.
             // TODO: the 0.95 threshold above is completely arbitrary and needs
             // adjustments.
             join_type = LineJoin::Miter
-        } else if join_type == LineJoin::Miter && normal.square_length() > limit * limit {
+        } else if join_type == LineJoin::Miter && self.miter_limit_is_exceeded(normal) {
             // Per SVG spec: If the stroke-miterlimit is exceeded, the line join
             // falls back to bevel.
             join_type = LineJoin::Bevel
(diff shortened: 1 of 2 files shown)


o0Ignition0o:
Hey @nical , feel free to tell me if:
- There's a better way to represent the svg specification than using consts 
- Panicking during the obj construction is fine by you, or if i should set default value instead
- I m allowed to link to the official svg spec in the comments ? 

Thaks a lot :D

o0Ignition0o:
` else if join_type == LineJoin::Miter && normal.square_length() > limit * limit {
            // Per SVG spec: If the stroke-miterlimit is exceeded, the line join
            // falls back to bevel.
            join_type = LineJoin::Bevel
        } ` Ok it looks like I m' reimplementing something you already have ^^' I'll refactore my code to use your version of "stroke-miterlimit is exceeded"

o0Ignition0o:
I guess I'm almost there, still need to handle a weird glitch on the top left corner: 
image
image
No idea whether it's related to your comment https://github.com/nical/lyon/pull/166#issuecomment-328593353 or not :/
BTW I'd love to add some tests to it, but I dont' really know If it's relevant, and how to proceed ^^'

tessellation/src/path_stroke.rs
```diff
@@ -746,6 +743,79 @@ impl<'l, Output: 'l + GeometryBuilder<Vertex>> StrokeBuilder<'l, Output> {
...
+    ) -> (VertexId, VertexId) {
+        let v = add_vertex!(
+                    self,
+                    Vertex {
+                        position: self.current,
+                        normal: normal,
+                        advancement: self.length,
+                        side: front_side,
+                    }
+                );
```
nical: nit: let's use block indentation here, like this:
```rust
        let v = add_vertex!(
            self,
            Vertex {
                position: self.current,
                normal: normal,
                advancement: self.length,
                side: front_side,
            }
        );
```

tessellation/src/path_stroke.rs
```diff
@@ -746,6 +743,79 @@ impl<'l, Output: 'l + GeometryBuilder<Vertex>> StrokeBuilder<'l, Output> {
...
+
+    fn tessellate_miter_clip_join(
+        &mut self,
+        prev_tangent: Vec2,
+        next_tangent: Vec2,
+        front_side: Side,
+        back_vertex: VertexId,
+        normal: Vec2,
+    ) -> (VertexId, VertexId) {
+        if self.miter_limit_is_exceeded(normal) {
```
nical: nit: really a minor thing but I tend to prefer when the complex code is out of the nested blocks, for example:

```rust
if condition {
    // complicated stuff
    return ...
}
return simple_stuff()
``` 

could be implemented like:

```rust
if !condition {
    return simple_stuff()
}
// complicated stuff
return ...
``` 

It helps with keeping the cyclomatic complexity low (what a scary word!) and and in my experience the code looks a tad less intimidating that way.

nical:
> BTW I'd love to add some tests to it, but I dont' really know If it's relevant, and how to proceed ^^'

Yeah, I have put a lot of focus on the fill tessellator so the latter has a pretty great test coverage but it would be good to start thinking about what kind of testing would be useful for the stroke tessellator.
I think that's more of a general thing, to look into outside the scope of this PR.

o0Ignition0o:
I have to do an other pass checking for unused variables + silly stuff that could be optimized away, but I feel like it should work fine :)

o0Ignition0o:
There might still be something wrong thoug, since I don't really understand the "neg_if_right" part of the bevel implementation, so I did not use it (and I might mix up start_vertex and end_vertex i guess?)

o0Ignition0o:
This should be it ! :) 
image

tessellation/src/path_stroke.rs
```diff
@@ -616,20 +615,21 @@ impl<'l, Output: 'l + GeometryBuilder<Vertex>> StrokeBuilder<'l, Output> {
...
-                        advancement: self.length,
-                        side: front_side,
-                    }
-                );
-                self.prev_normal = normal;
-
-                (v, v)
+                self.tessellate_miter_join(
+                    front_side,
+                    normal
```
nical: Oops actually there was a bug here: this should be front_normal instead of normal, I'll land a quick fix.

PR #195, opened 2017-10-25
Fix math in inflection_approximation_range

### LINKED ISSUE #194: CubicBezierSegment flattening
The flatten algorithm fails for some `CubicBezierSegment`s. Here is an example:

```rust
extern crate lyon;

use lyon::math::point;
use lyon::bezier::CubicBezierSegment;

fn main() {
    let s = CubicBezierSegment {
        from: point(0.0, 0.0),
        ctrl1: point(0.0, 0.0),
        ctrl2: point(50.0, 70.0),
        to: point(100.0, 100.0),
    };

    s.flattened_for_each(0.1, &mut |p| {
        println!("{}", p);
    });
    // output:
    // (0,0)
    // (100,100)
}
```

The important part is that we have an inflection point at the origin and that `to.x = to.y`.

I think the problem is in `inflection_approximation_range`. Some of the math there looks dubious to me. I will make a PR shortly.

pizzaiter:
Fixes #194 . I tried to explain the math directly in the code comments.

### DIFF
--- a/bezier/src/flatten_cubic.rs
+++ b/bezier/src/flatten_cubic.rs
@@ -288,31 +288,34 @@ fn inflection_approximation_range(
     bezier: &CubicBezierSegment,
     tolerance: f32,
 ) -> Option<f32> {
-    let ctrl21 = bezier.ctrl1 - bezier.from;
-    let ctrl41 = bezier.to - bezier.from;
+    // Transform the curve such that it starts at the origin.
+    let p1 = bezier.ctrl1 - bezier.from;
+    let p2 = bezier.ctrl2 - bezier.from;
+    let p3 = bezier.to - bezier.from;
 
-    if ctrl21.x.abs() < 1e-5 && ctrl21.y.abs() < 1e-5 {
-        // In this case s3 becomes lim[n->0] (ctrl41.x * n) / n - (ctrl41.y * n) / n = ctrl41.x - ctrl41.y.
-        let s3 = ctrl41.x - ctrl41.y;
+    // Thus, curve(t) = t^3 * (3*p1 - 3*p2 + p3) + t^2 * (-6*p1 + 3*p2) + t * (3*p1).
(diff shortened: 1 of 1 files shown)


bezier/src/flatten_cubic.rs
```diff
@@ -288,31 +288,34 @@ fn inflection_approximation_range(
...
-    if s3 == 0.0 {
-        // This means within the precision we have it can be approximated
-        // infinitely by a linear segment. Deal with this by specifying the
-        // approximation range as extending beyond the entire curve.
-        return None;
+    } else {
+        // In this case p1 is away from zero.
+        // Choose the line in direction p1 and use that p1 and p2 are parallel.
+        s3 = p1.cross(p3) / p1.length();
     }
```
nical: I think that we should still handle the case where s3 is equal to 0.0.

pizzaiter:
As far as I understand, all comparisons with a NaN evaluate to `false`. So the function would return `None` in this case anyway. It will raise a floating point exception though, if that is a concern.

I will add a test with the curve from the issue.

nical:
> As far as I understand, all comparisons with a NaN evaluate to `false`

I thought so too, but I checked with the rust playground to be sure and it looks like it depends on the sign of the NaN, see: https://playtest.rust-lang.org/?gist=25daf84826700b7985e957374aeb8f3f&version=stable

pizzaiter:
You are generating `+inf` and `-inf` in the playground but no `NaN`. You would get a `NaN` from an expression like `0.0 / 0.0`.
https://playtest.rust-lang.org/?gist=bdc42a3dcb6b447f5d0b31684f61ec50&version=stable

It should not be possible for `tf` to be `-inf` because of the `.abs()` and `+inf < 1.0` evaluates to `false`.

nical:
Oops, you are correct. Thanks!

PR #215, opened 2017-10-31
WIP: Add svg_viewer example

whmountains:
This PR adds a CLI that renders SVG documents into a window with panning and zooming.

Why:

* I want to learn how to do it.
* It would be useful for future users who want to learn to use Lyon.

Right now it just renders the first path it finds in the file.  Gray stroke, no fill.


examples/svg_viewer/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+authors = ["whmountains <8680147+whmountains@users.noreply.github.com>"]
+name = "circuit_design"
```
zen3ger: shouldn't it be called svg_viewer?

whmountains: Hehe, good catch.

nical:
It's really cool that you are working on this, I'm eager to see how far we can push it! SVG is a big and complicated specification, but a lot of people have expressed interest in rendering SVG documents with lyon, so this should be useful if only as a reference.

I don't have time to give this a thorough look right now but I will come back to it asap.
I did see that you are using an xml parsing crate and I suggest switching to the svgparser crate instead. It's really good and provides everything needed for SVG, including parsing CSS attributes, etc. It's also the crate used by lyon_svg to parse the path syntax (although lyon_svg doesn't handle anything else at the moment).

whmountains:
I'm glad to hear you like the idea.

I will switch to the `svgparser` crate.  Thanks for the suggestion!  Feedback in general is always welcome because I'm relatively new to Rust.

examples/svg_viewer/src/main.rs
```diff
@@ -0,0 +1,334 @@
...
+    fill: Color
+}
+impl VertexConstructor<tessellation::FillVertex, GpuFillVertex> for VertexCtor {
+    fn new_vertex(&mut self, vertex: tessellation::FillVertex) -> GpuFillVertex {
+        assert!(!vertex.position.x.is_nan());
+        assert!(!vertex.position.y.is_nan());
+        GpuFillVertex {
+            // (ugly hack) tweak the vertext position so that the logo fits roughly
+            // within the (-1.0, 1.0) range.
+            position: (vertex.position * 0.0145 - vec2(1.0, 1.0)).to_array(),
```
nical: We should probably get rid of this hack here and use pan/zoom functionalities to navigate the viewer instead.

examples/svg_viewer/src/main.rs
```diff
@@ -0,0 +1,334 @@
...
+            // start of new tag
+            Token::SvgElementStart(tag) => {
+                last_tag = Some(tag);
+            }
+            // start of an SVG attribute
+            Token::SvgAttribute(name, value) => {
+                // ignore non-paths
+                if last_tag == Some(ElementId::Path) {
+                    match name {
+                        AttributeId::D => current_path.d = String::from(value.slice()),
```
nical: Not very important but you might as well just convert the string to a path directly here, and just manipulate path objects later.

examples/svg_viewer/src/main.rs
```diff
@@ -0,0 +1,334 @@
...
+    file.read_to_string(&mut input_buffer)
+        .expect("Error reading input file!");
+
+    // process the SVG tokens and generate an array of paths
+    let mut last_tag = None;
+    let mut svg_paths = Vec::new();
+    let mut current_path = SvgPath::default();
+    let mut tokens = Tokenizer::from_str(&input_buffer).tokens();
+    for token in &mut tokens {
+        match token {
```
nical: A lot of SVG drawing software like inkscape tend to apply transforms to most elements instead of specifying paths in the reference coordinate system of the document. It would be great to support parsing transform attributes here and either apply the transform in the VertexConstructor, or in the vertex shader.

whmountains:
Thanks for the review!  I'll address those points when I have a chance.

> maybe first get it to a point where it renders simple SVG files with fills, stokes, the stroke options that are already supported by the tessellator and transforms.

Sounds great!

examples/svg_viewer/src/main.rs
whmountains: Nice one!  Converting to a path immediately also avoids allocating another `String` because the path builder takes a slice.

examples/svg_viewer/src/main.rs
whmountains: Thanks for pointing that out.  I'm not super familiar with SVG transforms but I'll look into it.

RazrFalcon:
I think it's better to use [resvg](https://github.com/RazrFalcon/libresvg) instead of `svgparser`. It will return a rendering tree with all styles resolved, so all you need is to pass it to `lyon`.

nical:
Closing in favor of #336 which was built on top of this and has been merged.

Issue #206, opened 2017-10-29
Print the coordinates of the cursort when clicking in the "show" command of the app. 


nical:
It'd be a very useful tool when debugging.

The implementation of the show command is in `cli/src/show.rs`.

EloD10:
I take ;)

PR #219, opened 2017-11-02
Print the coordinates of the cursor when clicking in the window on the show command cli app and gfx_advanced example

EloD10:
#206

### DIFF
--- a/examples/gfx_advanced/src/main.rs
+++ b/examples/gfx_advanced/src/main.rs
@@ -358,6 +358,7 @@ fn main() {
         stroke_width: 0.0,
         target_stroke_width: 0.5,
         draw_background: true,
+        cursor_position: (0.0, 0.0)
     };
 
     let mut cmd_queue: gfx::Encoder<_, _> = factory.create_command_buffer().into();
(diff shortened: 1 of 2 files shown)


examples/gfx_advanced/src/main.rs
```diff
@@ -517,14 +519,37 @@ fn update_inputs(events_loop: &mut glutin::EventsLoop, scene: &mut SceneParams)
...
         match event {
             Event::WindowEvent {event: glutin::WindowEvent::Closed, ..} => {
                 println!("Window Closed!");
                 status = false;
-            },
-            Event::WindowEvent {event: glutin::WindowEvent::KeyboardInput {input: glutin::KeyboardInput {state: Pressed, virtual_keycode: Some(key), ..}, ..}, ..} => {
+            },           
+            Event::WindowEvent {
+                event: WindowEvent::MouseInput {
+                    state: ElementState::Pressed, button: glutin::MouseButton::Left,
```
nical: Almost there. Travis ci appears to complain about the `ElementState` symbol not being imported. Since `ElementState::Pressed` is imported you can just write `Pressed` here to get it to compile.

nical:
Thanks!

PR #241, opened 2017-11-11
Refactor the fill tessellator in preparation for more fill rules.

nical:
Keeping track of the winding numbers in the sweep line to implement more fill rules requires quite a bit of refactoring.


nical:
I am running into some difficulties with parts of the algorithm that are there to mitigate precision bugs but only work with the even-odd fill rule. Removing them means some tests break and it will take a little while until I figure out how to work around these precision issues in a way that works with more fill rules.
In the mean time it would be good to land this since there are a few things that are good to take in these refactorings, regardless of the non-zero fill rule.

PR #263, opened 2017-12-06
Implement a monotonic curve intersection algorithm.

nical:
The algorithm is adapted from pathfinder's binary search approach with a few enhancements:
 - solve_t_for_x takes a range of t to search into which makes it converge to the solution faster,
 - the main intersection loop only computes the mid point and reuse the result in future iterations (where the original algorithm re-computes the min and max points each time).

Unfortunately this algorithm makes the assumption that there can be at most one intersection between two monotonic segments (in fact there can be up to two) and may miss intersections when that's not the case. That said, this bug only seem to be possible when the curves cross twice in a way that should not mess up with the sweep line's invariant (it will just produce a few unexpectedly overlapping triangles in rare cases).

I think that the binary search can be made to converge faster by biasing the mid point towards the min or max depending on which has the smallest difference in y.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -361,6 +361,14 @@ dependencies = [
 ]
 
 [[package]]
+name = "geom_bench"
+version = "0.0.1"
+dependencies = [
+ "bencher 0.1.4 (registry+https://github.com/rust-lang/crates.io-index)",
+ "lyon 0.8.6",
+]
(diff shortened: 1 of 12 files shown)


nical:
Partially addresses #8 (we still need to implement some glue for non-monotonic curves).

Issue #143, opened 2017-08-16
Add an example in the doc of lyon_svg::parser::build_path


nical:
See: https://docs.rs/lyon_svg/0.7.0/lyon_svg/parser/fn.build_path.html

It would be helpful to have an example illustrating how this function is used and what it is useful for.

nical:
Fixed by @o0Ignition0o in #266.

PR #266, opened 2017-12-18
Added documentation to the build_path method.

o0Ignition0o:
This should handle #143 I guess.

Would be really eager to know if there's a way for me to preview the documentation :) 

Maybe I should add a link to the cli README.md file, or provide more information on the commands string I use ?

### DIFF
--- a/svg/src/parser/path.rs
+++ b/svg/src/parser/path.rs
@@ -23,7 +23,7 @@ use super::error::ParserError;
 /// // Create a simple path.
 /// let commands = &"M 0 0 L 10 0  10 10 L 0 10 z";
 /// let svg_builder = Path::builder().with_svg();
-/// let path = build_path(svg_builder, &commands);
+/// let path = build_path(svg_builder, commands);
 /// # }
 /// ```
 pub fn build_path<Builder>(mut builder: Builder, src: &str) -> Result<Builder::PathType, ParserError>
--- a/svg/src/parser/path.rs
+++ b/svg/src/parser/path.rs
@@ -21,7 +21,7 @@ use super::error::ParserError;
 /// # use svg::parser::build_path;
 /// # fn main() {
 /// // Create a simple path.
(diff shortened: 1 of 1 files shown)


svg/src/parser/path.rs
```diff
@@ -6,6 +6,26 @@ use path::math::{Point, point, Vector, vector, Radians};
...
+/// # Examples
+///
+/// ```
+/// # extern crate lyon_svg as svg;
+/// # extern crate lyon_path;
+/// # use lyon_path::default::Path;
+/// # use svg::parser::build_path;
+/// # fn main() {
+/// // Create a simple path.
+/// let commands = String::from("M 0 0 L 10 0  10 10 L 0 10 z");
```
nical: Nit: you don't need to build a string here, just using &"M..." should do.

o0Ignition0o: Fixed :D

svg/src/parser/path.rs
```diff
@@ -6,6 +6,26 @@ use path::math::{Point, point, Vector, vector, Radians};
...
+/// ```
+/// # extern crate lyon_svg as svg;
+/// # extern crate lyon_path;
+/// # use lyon_path::default::Path;
+/// # use svg::parser::build_path;
+/// # fn main() {
+/// // Create a simple path.
+/// let commands = &"M 0 0 L 10 0  10 10 L 0 10 z";
+/// let svg_builder = Path::builder().with_svg();
+/// let path = build_path(svg_builder, &commands);
```
nical: nittiest nit: not a big deal but you are passing a `&&str` which gets auto-dereffed into a `&str`. you can just pass `commands` without the `&`.
I am a bit pickier than I should but since it is documentation, I'd rather make sure the example doesn't make it look like `build_path` takes a `&&str`

o0Ignition0o: Please be as picky as you can ! I know you don't mean to be rude, and I want to improve as much as possible ! :)

nical:
Thanks!

Issue #193, opened 2017-10-23
Add testing to the Stroke Tessellator


o0Ignition0o:
Followup on this https://github.com/nical/lyon/pull/192#issuecomment-338504066 it would be nice to add some tests to the tessellator.

Before trying to implement this, I'd love to bikeshed on what would be best to test the whole module.

- Unit tests ? 
- Fuzz tests ?
- Maybe some reftests as used on webrender or servo ? ( https://github.com/servo/webrender/commit/910f58253f5fa236dde04ea277c9d4b1e818dcb2 )

I don't know who to cc on this, so feel free to add cc's and comments :)

nical:
> Fuzz tests ?

Fuzzing is already in place (from the cli directory, run `cargo run -- fuzz -s`), but it isn't very useful in its current state because there isn't a lot of assertions in the stroke tessellator for the fuzzer to find.

Reftests would be pretty great. It would pretty cool if the cli program had a sub-command to render fills and strokes to an image. We could build a script on top of that to compare the images and as a bonus log the errors using the same format as gecko/webrender to reuse the reftest analyser tool.

PR #267, opened 2017-12-22
[WIP] Export command

o0Ignition0o:
Created an export command, which will allow to render a path and export it to a PNG or JPG File.
Still have to implement it though :). 

This is one of the required steps to allow usage reftests later, as described in #193

### DIFF
--- a/cli/src/export.rs
+++ b/cli/src/export.rs
@@ -24,7 +24,7 @@ pub struct FileFormatParseError;
 
 pub enum FileFormat {
     JPG,
-    PNG
+    PNG,
 }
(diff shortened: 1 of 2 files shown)


o0Ignition0o:
Tryouts to add a render method. It kind of displays fine, and I'm able to compute the window size, but I can't adjust the scene according to it. I'll figure it out later.

Once I found a way to handle this, I need to be able to render only one frame (maybe removing the while loop and executing it only once) and save the buffer to a BMP / PNG file.
Glad to see I'm moving forward though :')

PR #279, opened 2017-12-29
Provide both point and "t" intersection interfaces.

nical:


geom/src/quadratic_bezier.rs
```diff
@@ -572,7 +597,7 @@ fn monotonic_solve_t_for_x() {
     for i in 0..10u32 {
         let t = i as f32 / 10.0;
         let p = curve.sample(t);
-        let t2 = curve.assume_monotonic().solve_t_for_x(p.x, 0.0..1.0, tolerance);
```
o0Ignition0o: I don't understand what happened to the tolerance here 😅

nical: The monotonic quadratic bézier implementation of `solve_t_for_x` doesn't need a tolerance (nor a narrowed range of t). The generic monotonic implementation of that function used with cubic curves does, though.

o0Ignition0o: Oh that's why, thank's a lot for your explanation :)

PR #280, opened 2017-12-29
Make geom generic over Float

pizzaiter:
This PR makes the geom crate generic over Float (plus some other traits when needed).

There is one small FIXME in the code, but this will need changes in euclid. Also, I removed util::fast_atan2() and used the version from euclid::Trig.


pizzaiter:
:smile: :tada:

PR #285, opened 2018-01-01
Add option --precision to CLI tessellate command

SilentByte:
Allows to specifiy the output floating point precision.
If no precision flag is specified, the output is identical
to the previous version.

### DIFF
--- a/cli/src/commands.rs
+++ b/cli/src/commands.rs
@@ -7,7 +7,7 @@ pub struct TessellateCmd {
     pub path: Path,
     pub fill: Option<FillOptions>,
     pub stroke: Option<StrokeOptions>,
-    pub precision: Option<usize>,
+    pub float_precision: Option<usize>,
 }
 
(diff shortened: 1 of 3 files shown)


SilentByte:
The main incentive for this was that the output of the tessellate command isn't compatible with Rust if values happen to be integers, so the resulting array of tuples couldn't be compiled. I figured a precision option would be a good solution that also adds additional benefits.

nical:
Good idea! Thanks for this PR. "Precision" can be a bit confusing due to other precision parameters in lyon's algorithms like the flattening tolerance threshold, could you rename the command-line flag into `--float-precision` and the TessellateCmd member into `float_precision`?

SilentByte:
Thanks, good points!

Changed --precision to --float-precision and removed short flag -p (now only available via --float-precision).

nical:
Looks like travis is choking on itself. The PR looks good though, let's merge!

PR #299, opened 2018-01-13
Remove unnecessary ApproxEq bounds

pizzaiter:
Depends on servo/euclid#259 and needs to wait for a new euclid version.


nical:
[Euclid 0.16.2](https://crates.io/crates/euclid/0.16.2) is out.

pizzaiter:
I bumped the euclid version. Should be good to merge now.

PR #329, opened 2018-02-11
Fix + update examples

whmountains:
### Fixes

* Accessing `use lyon::extra` now requires a feature flag in `Cargo.toml`
* `in vector a_position;` on line 155 of `examples/gfx_basic/src/main.rs` needed to be changed to `in vec2 a_position;`.  I don't know enough about GLSL to say if this was a typo or what, but I had to make that change to make the shader compile.

### Upgrades

I took the opportunity to upgrade all the dependencies to their latest versions.  I had to change the code in a few places where the `glutin` API had changed.  I used `glutin@0.12.0` because `glutin@0.12.1` depends on `winnit@0.10.1` which has been yanked from crates.io.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -796,10 +796,10 @@ dependencies = [
 name = "intersections example"
 version = "0.1.0"
 dependencies = [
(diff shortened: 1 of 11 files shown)


nical:
Thanks!

PR #334, opened 2018-02-26
New fill tessellator

nical:
A rewrite of the fill tessellator with a few major differences:

 - it doesn't use fixed point coordinates internally,
 - ~it understands quadratic bézier curves and lines instead of just lines (if this works out the tessellator will be resolution-independent)`~ (Edit: backed down on handling curves in the hope of being able to ship some day!),
 - the logic is designed to better handle various fill rules like non-zero (while the other tessellator was designed around the even-odd fill rule and retro-fitting non-zero is hard),
 - most importantly, the new algorithm is designed to be more robust against floating point precision issues.

It also shares some important similarities with the current tessellator:

 - also based on monotone decomposition (as opposed to using trapezoids like pathfinder),
 - runs in a single pass,
 - treats events in batches that are at the same position, which is the current tessellator's main particularity compared to other monotone tessellators out there and is what makes it feasible to run in single pass.

This is far from finished, but if it works out it'll  be pretty exciting.


nical:
Quick update. I have been working on this on and off (more off than on I have to admit). The main goal for the new tessellator is uncompromising robustness (I'll see about curves later), and in order to make sure the algorithm is able to recover from invalid states introduced by floating point inaccuracy, I have been working on making it pass the whole test suite without handling self-intersections. So in the many tests that do have self-intersections, the algorithm at some point gets into a pretty bad invalid state and must recover, which is actually pretty tough but I think will pay off in the long run.
Once I feel confident about the approach and the implementation, I'll of course add the code that detects self-intersections and split the edges accordingly.

I'm down to 8 tests failing (out of 133), and the new tessellator even passes three tests that the current one fails.

So even though it's a pretty old branch by now (Feb 2018, ouch!), I keep coming back to it when I have time and energy, and not all hope is lost.

nical:
I am getting close to merging the biggest pull request I ever made (and I hope to never beat this personal record of mine in the future).

PR #336, opened 2018-03-03
Add svg_render example

zen3ger:
Based on the suggestions in #215 I've made an `svg_render` example.
It renders both shapes and strokes.

I couldn't figure out how to pass stroke styles from `resvg` to `lyon`. If someone knows, I'd be happy to add that too.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,10 +50,11 @@ members = [
     "cli",
     "examples/gfx_advanced",
     "examples/gfx_basic",
-    "examples/intersections",
-    "examples/walk_path",
     "examples/glium_basic",
     "examples/glium_basic_shapes",
+    "examples/intersections",
+    "examples/svg_render",
+    "examples/walk_path",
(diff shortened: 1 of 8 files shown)


nical:
Nice! There are a couple of features that would be nice to add like having correct aspect ratio no matter the size of the window and automatically positioning the view to fit the document, but we can add them as followups.

Could you squash the commits? Also could you remove center.svg, test.svg is more useful for people trying out the example.

zen3ger:
I've fixed the centerring. By this week I'll probably fix the scaling too. It took quite some time to find a [gfx example](https://gist.github.com/matthewjberger/c1cd9406220f55f6fa07be78535744f8) wich uses a projection matrix...

All in all, few more days and I finish off this one.

zen3ger:
Done, I guess. Let me know if I should fixup something.

nical:
Sweet! Thanks a lot!

PR #338, opened 2018-03-19
Fix field access error in svg_render example

SilentByte:
Quick fix of the svg_render example to make it compile again.


nical:
Thanks!

PR #350, opened 2018-05-17
Use usvg instead of resvg, minor fixes

### LINKED ISSUE #349: Use usvg instead of resvg in the svg_render example
It's the same underlying code, resvg depends on usvg and we only use the usvg part so let's just depend on that crate instead.

zen3ger:
Typo and zoom control char fixes.
Added option to set msaa from commandline.
Replace resvg with usvg.

Closes #349

### DIFF
--- a/examples/svg_render/src/main.rs
+++ b/examples/svg_render/src/main.rs
@@ -1,21 +1,23 @@
+extern crate clap;
 #[macro_use]
 extern crate gfx;
 extern crate gfx_device_gl;
 extern crate gfx_window_glutin;
 extern crate glutin;
 extern crate lyon;
-extern crate resvg;
+extern crate usvg;
 
 mod path_convert;
 mod stroke_convert;
 mod render;
 
-use resvg::tree::Color;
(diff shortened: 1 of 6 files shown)


zen3ger:
I saw that the `cli` uses `clap`, I thought it might be an overkill here, but should I use it?

examples/svg_render/src/main.rs
```diff
@@ -240,10 +264,10 @@ fn update_inputs(scene: &mut Scene, event_loop: &mut glutin::EventsLoop) -> bool
                 VirtualKeyCode::Escape => {
                     status = false;
                 }
-                VirtualKeyCode::PageDown => {
```
nical: I'd prefer to keep PgUp/down to control the zoom like in the other examples and the cli app.

zen3ger:
Sorry, I saw the `println!()` with sqare brackets and thought it was messed up. I fixed the text and added `clap`. If everything is fine, I'll squash the commits.

PR #362, opened 2018-06-03
Add custom formatting to cli tessellate cmd

zen3ger:
### What's working:
Same output as before using the default format string.
Custom format strings which don't contain `{}` inside expansion blocks (so far) work as expected.
Escaping `{}` in format string with `\{` and `\}`.

### For the future:
- reduce string copies
- simplify syntax


zen3ger:
I think it worth mentioning that this can be **finally** done:
```
$ cargo run -- tessellate --format "@indices{sep=-}{Na} BATMAN!" "$TEST"
Na-Na-Na-Na-Na-Na-Na-Na-Na-Na-Na-Na-Na-Na-Na-Na-Na-Na-Na-Na-Na BATMAN!
```

zen3ger:
@nical if you have time give it a try, you might end up writing some formatter that will fail. Also do you think the doc is enough?

cli/README.md
```diff
@@ -58,6 +58,25 @@ indices: 6
...
+@vertices{sep=<item separator>}{<item format string>}
+
+```
+The `<item separator>` can be any string. In case of `@indices` the `<item format string>` can contain the `{i}` and `{index}` variable tags, while `@vertices` can have `{pos.x}`, `{pos.y}`, `{position.x}` and `{position.y}` variable tags.
+
+As an example, the default output format is the following:
+```
+vertices: [@vertices{sep=, }{({position.x}, {position.y})}]\nindices: [@indices{sep=, }{{index}}]
+```
+If you want to use `{` and `}` in the format string you have to write `\{` and `\}`, as normally `{}` groups mark a variable expension. Here's an example:
```
zen3ger: typo, "expansion"

nical:
I love it! I'm almost able to do a json output ( Running into issues with `\}`). It would be great to be able to do a bunch of other simple formats like .obj (for this one we might need to add a `@triangles{sep}{formatter}` that gives access to triplets of indices).

I'm happy to land this as is and keep iterating on it unless you prefer to do some more polish before it host the master branch.

zen3ger:
Ok, I added `@triangles` and found a way to do JSON like output without overcomplicating the regex patterns.
If the end of the iterator expander also marked with `@` than any trailing `\}` would be parsed in a separate block. I decided to add `{fmt=...}` for the item format string just to make it consistent with the separator syntax.
The syntax became `@indices{sep=, }{fmt={index}}@`...

A JSON like output with `@triangles`:
```
$> lyon tessellate --format '\{\n "triangles": \{\n@triangles{sep=,\n}{fmt=  "triangle": \{\n   "{0}",\n   "{1}",\n   "{2}"\n  \}}@\n \}\n\}' "M 0 0 L 10 0  10 10 L 0 10 z"
{
 "triangles": {
  "triangle": {
   "1",
   "0",
   "2"
  },
  "triangle": {
   "1",
   "2",
   "3"
  }
 }
}
```

nical:
Nice! Why is the trailing `@` needed if non-special `}` are also escaped?

zen3ger:
`\}` and `\{` was "escaped" via regex pattern. The `\` is then removed before print. Unfortunately I couldn't figure out how **not** to match on `\}` chars after the `{formater}` block, ie. `\{@indices{sep=,}{\{{index}\}}\}` block, as the formater capture `\{(.*?)\}` would match until the first variable closing (`\{{index` and then `}\}}\}` is the left over string) and `\{(.*)\}` would match until the very last `}`.

Basically, `@` used to split the formater into substrings for parsing so `@indices{}{} @vertices{}{}` is `["indices{}{}", " ","vertices{}{}"]` and than matchin is done... By adding `@` after the formatter any trailing `\}` would be in a different string which would be just pushed to the buffer without var expansion. Leading `\{` are solved also via `@` before the formater, and I fixed the varibale match regex to capture `\{` before the var pattern.

If the `@` after the formatter is anoying I'll come up with something better.

nical:
I see. It'd be nice to come back and simplify the syntax eventually, but the feature is already super cool in its current state so I'm happy to merge this as is (I'd merge it right now but there is still [wip] in the title so I am not sure whether you want to add more stuff).

Issue #327, opened 2018-02-09
lyon_geom: Calculating path's bounding box with stroke


RazrFalcon:
I'm trying to find path's bounding box including stroke, but can't find a way to do this.

Currently, I'm using just a `lyon_geom::CubicBezierSegment::bounding_rect`, but it obviously doesn't count the stroke.

Is there are existing tools in lyon for this? Or does they planned?

nical:
There isn't anything in lyon to compute the bounding box of a path at the moment. I think that it would be useful to add to lyon, at least for fills.

For strokes the main question is whether you are ok with an approximation. Inflating the path's bounding box with the stroke width should be an okay-ish approximation in most cases except for miter joins that might overflow that rectangle a little bit. inflating the bounding box a little more using the miter limit should produce a conservative (slightly too large but guaranteed to contain the whole stroke) approximation.

Otherwise if you need an exact stroke bounding box, you'll have to loop over all commands, compute the joins to take miters that could extend far away into account and get the precise bounding box that way. I think that this would be a bit too specific for lyon, though.

RazrFalcon:
I see. The `cairo` library supports this, kinda ([actual function](https://github.molgen.mpg.de/git-mirror/cairo/blob/master/src/cairo-stroke-style.c#L97..L128)). But looks like they also use an approximation.

Currently, I'm just expanding the bbox by `stroke-width/2.0`. It almost works, but not when we have acute angles. Example:
![out3](https://user-images.githubusercontent.com/725494/36040529-6160bc7a-0dce-11e8-9119-321c9e8e962e.png)

>There isn't anything in lyon to compute the bounding box of a path at the moment.

Why not? All I need is `*Segment::bounding_rect`. Yes, there is some manual manipulations, but still bearable.

nical:
> Why not?

There is no particular reason not to do it, it's simply that the code hasn't been written yet. Now that you bring a use case for it I have a motivation to add it.

nical:
Path bounding boxes are [available in 0.11](https://docs.rs/lyon_algorithms/0.11.1/lyon_algorithms/aabb/index.html).

RazrFalcon:
Great! Thanks.

PR #376, opened 2018-06-25
Arc and path bounding rects.

nical:
The initial goal was simply to implement fast and tight axis-aligned bounding rects for paths but it turned into an odyssey of implementing tight bounding rects for arcs and finding/fixing arc bugs.


nical:
The added function `lyon_algorithms::aabb::bounding_rect` mostly addresses #327 (although the stroke width isn't taken into account).

PR #389, opened 2018-07-10
Bump euclid to 0.18.2 for lyon-geom

BorisChiou:


BorisChiou:
Sorry, I still need to update lyon-path.

nical:
Looks good but you don't actually need to do this, because lyon_geom 0.11.0 already depends on euclid 0.18.0, you don't have to manually update the last number to 0.18.2. If two crates depend on 0.18.x, cargo will consider that they are compatible and will use the highest version of x requested.

BorisChiou:
Cool. OK. So I only need to update pathfinder to use a newer lyon-geom and lyon-path. Thanks.

PR #419, opened 2018-10-05
 Add arc_to_cubic_beziers for Arcs. 

kleintom:
Note that I increased the step for cubic beziers to pi/2 from the pi/4 used in the quadratic case - I'm new to this so I'm not sure if that's appropriate or not (it's based on a comment at https://pomax.github.io/bezierinfo/#circles_cubic).

I'm attaching plots of the four test Arcs from arc.rs with their cubic approximations, plus a fifth with a larger sweep_angle
[graphs.pdf](https://github.com/nical/lyon/files/2450677/graphs.pdf)

### DIFF
--- a/geom/src/arc.rs
+++ b/geom/src/arc.rs
@@ -9,6 +9,7 @@ use generic_math::{Point, point, Vector, vector, Rotation2D, Transform2D, Angle,
 use segment::{Segment, FlattenedForEach, FlatteningStep, BoundingRect};
 use segment;
 use QuadraticBezierSegment;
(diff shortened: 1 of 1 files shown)


nical:
Splendid!

> Note that I increased the step for cubic beziers to pi/2 from the pi/4 used in the quadratic case

Yes that's appropriate, cubic béziers are able to fairly closely approximate a quarter of an ellipse.
At a glance the code looks good, I'll give it a closer look before tomorrow.

Thanks a lot!

PR #420, opened 2018-10-05
Update Cargo.lock for svg_render's name change.

kleintom:
Each time I build cargo makes this update - looks like it got missed in #417?


nical:
Ah yes indeed, thanks!

PR #421, opened 2018-10-05
lyon_geom version 0.12.2.

nical:
Publish the new arc to cubic bézier conversion added in #419.


nical:
https://crates.io/crates/lyon_geom/0.12.2

PR #424, opened 2018-10-18
Changes around cubic bezier line intersections

kleintom:
(I haven't checked the behavior when a curve and a line segment intersect at an endpoint: I _think_ the old behavior depended on the "left/top closed, right/bottom open" behavior of Rect.contains, and maybe rounding issues; now I think we have a "left closed, right closed" behavior (and still rounding issues?), so we might include an endpoint intersection in some cases where we didn't previously.  Maybe a new issue? or I can address it here if you like.)

### DIFF
--- a/geom/src/cubic_bezier.rs
+++ b/geom/src/cubic_bezier.rs
@@ -612,16 +612,29 @@ impl<S: Scalar> CubicBezierSegment<S> {
         }
 
         let intersections = self.line_intersections_t(&segment.to_line());
-        let aabb = segment.bounding_rect();
 
         let mut result = ArrayVec::new();
+        if intersections.len() == 0 {
+            return result;
+        }
+
+        let seg_is_mostly_vertical =
+            S::abs(segment.from.y - segment.to.y) >= S::abs(segment.from.x - segment.to.x);
+        let (seg_long_axis_min, seg_long_axis_max) = if seg_is_mostly_vertical {
+            segment.bounding_range_y()
+        } else {
+            segment.bounding_range_x()
+        };
+
         for t in intersections {
-            if aabb.contains(&self.sample(t)) {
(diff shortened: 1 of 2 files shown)


kleintom:
> I haven't checked the behavior when a curve and a line segment intersect at an endpoint

(Ah, `line_intersections_t` excludes `t = 0` and `t = 1`, so that already excludes curve/line segment endpoint/endpoint intersections.  You could still get line segment endpoint intersecting curve interior I guess, but not the other way around.)

geom/src/utils.rs
```diff
@@ -55,13 +55,13 @@ pub fn cubic_polynomial_roots<S: Scalar>(a: S, b: S, c: S, d: S) -> ArrayVec<[S;
 
     if S::abs(a) < S::EPSILON {
         // quadratic equation
-        let delta = b * b - S::FOUR * a * c;
+        let delta = c * c - S::FOUR * b * d;
```
nical: Oops!

geom/src/cubic_bezier.rs
```diff
@@ -612,16 +612,37 @@ impl<S: Scalar> CubicBezierSegment<S> {
...
+        } else {
+            segment.bounding_range_x()
+        };
+
         for t in intersections {
-            if aabb.contains(&self.sample(t)) {
-                let t2 = (self.sample(t) - segment.from).length() / segment.length();
-                result.push((t,t2));
+            if seg_is_mostly_vertical {
+                let intersection_y = self.y(t);
```
nical: Nit: How about

```rust
let intersection_xy = if seg.is_mostly_vertical { self.y(t) } else { self.x(t) };
if intersection_xy >= seg_long_axis_min && intersection_xy <= seg_long_axis_max {
   let t2 = (self.sample(t) - segment.from).length() / segment.length();
   result.push((t, t2));
}
```

nical:
Looks like this needs to be rebased and then we're good to go.

PR #425, opened 2018-10-18
Update Cargo.lock for the recent lyon_geom version bump.

kleintom:
Cargo makes these changes each time I build.


nical:
Thanks!

PR #426, opened 2018-10-21
Don't divide by zero in cubic_bezier::split_range.

kleintom:
We only actually divide by zero and return NANs when t_range == 1.0..1.0, but we
might as well short circuit the entire t_range.start == t_range.end case while
we're at it.

I ran into this while testing the cubic bezier intersection work: the domain of one of the curves was reduced by fat-lining to 1.0..1.0 because their was an endpoint intersection there, and then I was seeing a lot of NANs after trying to split the original curve down to that new interval.


nical:
Looks good, thanks!

PR #428, opened 2018-12-18
Fix vignette in gfx_advanced example

gralpli:
The clamp function is defined as `float clamp(float x, float minVal, float maxVal)`, so the first argument should be the value to be clamped. It was errorneously given as the last argument.


nical:
Ah, yes indeed, thanks!

PR #432, opened 2018-12-27
Fix `serialization` feature of lyon_path

SimonSapin:
Without this I get build errors like below. Maybe something should be added to `.travis.yml`?

```rust
$ cargo build --all-features
error[E0277]: the trait bound `VertexId: serde::Serialize` is not satisfied
  --> path/src/default.rs:31:5
   |
31 |     vertex: VertexId,
   |     ^^^^^^ the trait `serde::Serialize` is not implemented for `VertexId`
   |
   = note: required by `serde::ser::SerializeStruct::serialize_field`
```


nical:
Oops, thanks!

PR #433, opened 2019-01-02
Intersection detection adjustments for cubic  bézier curves

kleintom:
Comments are in the commit messages.

The numbers I'm getting from my bench runs have changed somewhat from previous runs (I don't know why), but fwiw, without these patches, I get:
f32 curves: 22,552,339
f64 curves: 25,775,672

With the epsilon change (the first commit), I get:
f32 curves: 22,628,392 (expected no change from the previous f32 bench)
f64 curves: 24,613,664 (expected faster than the previous)

And with the new f32 sampling code (the second commit), I get:
f32 curves: 22,874,551 (expected slower than the previous)
f64 curves: 24,535,376 (expected more or less no change from the previous)

### DIFF
--- a/geom/src/cubic_bezier_intersections.rs
+++ b/geom/src/cubic_bezier_intersections.rs
@@ -259,7 +259,7 @@ fn add_curve_intersections<S: Scalar>(
         return call_count;
     }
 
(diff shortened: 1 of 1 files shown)


nical:
Thanks!

PR #456, opened 2019-02-21
Fix Triangle.contains_point impl to not care about vertex winding order

hardiesoft:
It seems like vertex winding order of triangles should not matter when checking to see if a point is inside a triangle - especially in 2D.  If it was intentional behaviour (for cases when you want to ignore the hit-test when the triangle is not front-facing?) then please document the intended usage.

Thanks!
Jon


nical:
It was indeed a bug. Thanks a lot!

PR #463, opened 2019-04-07
Fix `math::F64Point` documentation

demurgos:
This commit fixes a typo in the `math::F64Point` documentation. This type is an alias for `euclid::Point2D<f64>` (not `euclid::Point2D<f32>`).


nical:
Woops, thanks!

PR #464, opened 2019-05-01
New path data structures and iterators

nical:
This is another massive refactoring of the iterator API:

PathEvent is now generic:

```rust
pub enum PathEvent<Endpoint, CtrlPoint> {
    Line { from: Endpoint, to: Endpoint },
    Quadratic { from: Endpoint, ctrl: CtrlPoint, to: Endpoint },
    Cubic { from: Endpoint, ctrl1: CtrlPoint, ctrl2: CtrlPoint, to: Endpoint },
    Begin { at: Endpoint, },
    End { last: Endpoint, first: Endpoint, close: bool },
}
```

And iterators are responsible for *always* emitting an `End` event at the end of each sub-path.

In addition a few new data structures are introduced:

- `PathCommands` encodes path events with ids for endpoints and control points so that the user can manage the layout and storage of themselves the vertices themselves.
- `PolygonSlice` and `IdPolygonSlice` provide similar fenctionality to deal with simple polygons easily and efficiently.

The general idea is that complex use cases can be expressed with `PathCommands` and `IdPathSlice` at the cost of some extra overhead, while for simple and common use cases, `Path`/`PathSlice`, and the polygon types offer faster/simpler APIs.

In the process of making this change I removed a lot of functionality that's been around but not used much such as `QuadraticEvent` and iterator adapters to convert from path events to svg events.


nical:
This got folded into the new-tess branch.

PR #477, opened 2019-06-20
Path Stroke Tessellator Improvements to handle overlaps

Dollab:
I've modified the path stroke tessellator to better handle specific cases such as acute angles or very short edges that would induce the stroke of two consecutive edges to overlap.

The modifications were designed to change as little as possible the original code at the expense of perfect geometry. In particular, without a second pass it seems impossible to generate a perfect tessellation as we need the information on the geometry of next joins that might overlap.

The solution proposed works with some overlapping triangles and collinear `flat` triangles. A few vertices were also added in some specific joins that needed two distincts vertices on the back of the join.

I've also added an Enum called Order, to specify whether an overlaps is occurring `before` or `after` the current join.

### DIFF
--- a/tessellation/src/lib.rs
+++ b/tessellation/src/lib.rs
@@ -261,6 +261,27 @@ impl Side {
     pub fn is_right(self) -> bool { self == Side::Right }
 }
 
+/// Before or After. Used to describe position relative to a join.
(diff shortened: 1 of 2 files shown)


nical:
Thanks a lot! I'll  have a look as soon as I can.

nical:
Thanks!

Issue #481, opened 2019-07-13
Path fails to tesselate


Yatekii:
Hi!

I found another path that does not tesselate:

```
Path {
    points: [
        (0.9177246,0.22070313),
        (0.9111328,0.21826172),
        (0.91625977,0.22265625),
        (0.9111328,0.22753906),
        (0.9309082,0.2397461),
        (0.92163086,0.24121094),
        (0.91796875,0.23486328),
        (0.91845703,0.23999023),
        (0.90649414,0.24633789),
        (0.9038086,0.23022461),
        (0.89575195,0.23779297),
        (0.88671875,0.23583984),
        (0.88427734,0.2277832),
        (0.88671875,0.22143555),
        (0.8964844,0.21972656),
        (0.904541,0.22460938),
        (0.9111328,0.21459961),
        (0.907959,0.24072266),
        (0.9094238,0.24169922),
        (0.9104004,0.24047852),
        (0.9111328,0.23950195),
        (0.91674805,0.24047852),
        (0.91259766,0.23803711),
        (0.8864746,0.22998047),
        (0.88793945,0.22998047),
        (0.8874512,0.22827148),
        (0.8852539,0.2265625),
        (0.8864746,0.22924805),
        (0.8869629,0.22607422),
        (0.88793945,0.22827148),
        (0.8894043,0.22729492),
        (0.8869629,0.22607422),
        (0.8918457,0.22680664),
        (0.89453125,0.2265625),
        (0.89282227,0.22558594),
        (0.8911133,0.2241211),
        (0.8898926,0.22436523),
        (0.89038086,0.22558594),
        (0.9238281,0.23022461),
        (0.9213867,0.23022461),
        (0.91918945,0.22729492),
        (0.92211914,0.22680664),
    ],
    verbs: [
        MoveTo,
        LineTo,
        LineTo,
        LineTo,
        LineTo,
        LineTo,
        LineTo,
        LineTo,
        LineTo,
        LineTo,
        LineTo,
        LineTo,
        LineTo,
        LineTo,
        LineTo,
        LineTo,
        LineTo,
        Close,
        MoveTo,
        LineTo,
        LineTo,
        Close,
        MoveTo,
        LineTo,
        LineTo,
        Close,
        MoveTo,
        LineTo,
        LineTo,
        Close,
        MoveTo,
        LineTo,
        LineTo,
        Close,
        MoveTo,
        LineTo,
        LineTo,
        Close,
        MoveTo,
        LineTo,
[...]

Issue #476, opened 2019-06-18
Fail to build a path from a builder (E01).


Yatekii:
I run into: `Failed to tesselate path.: Internal(E01)`.

The Builder contains those points:

```
(10720.101,7120.1816)
(10720.099,7120.1816)
(10720.1,7120.182)
(10720.099,7120.1836)
(10720.101,7120.1846)
(10720.098,7120.1855)
(10720.096,7120.189)
(10720.096,7120.1885)
(10720.094,7120.188)
(10720.095,7120.1885)
(10720.095,7120.1885)
(10720.094,7120.189)
(10720.095,7120.1885)
(10720.091,7120.1865)
(10720.096,7120.1855)
(10720.097,7120.1836)
(10720.098,7120.1846)
(10720.099,7120.1816)
(10720.098,7120.1826)
(10720.097,7120.181)
(10720.1,7120.1807)
```
The First is a `MoveTo` and the others are a `LineTo`. It also contains a `Close` verb at the end.

Yatekii:
Update: I was wrong: It happens in the FillTesselator. The Path that causes the issue is this:

```
Path {
    points: [
        (2439.4155,1620.2511),
        (2439.4153,1620.2509),
        (2439.4155,1620.2507),
        (2439.4153,1620.2509),
        (2439.4153,1620.2507),
        (2439.415,1620.2509),
        (2439.4153,1620.2512),
        (2439.415,1620.2516),
        (2439.4148,1620.2517),
        (2439.4143,1620.2515),
        (2439.4136,1620.2515),
        (2439.4133,1620.2511),
        (2439.4126,1620.2507),
        (2439.4126,1620.2506),
        (2439.4128,1620.2506),
        (2439.4128,1620.2502),
        (2439.413,1620.2501),
        (2439.413,1620.2498),
        (2439.4136,1620.2495),
        (2439.4148,1620.25),
        (2439.4148,1620.2499),
        (2439.415,1620.25),
        (2439.4153,1620.2498),
        (2439.415,1620.2499),
        (2439.415,1620.2496),
        (2439.4155,1620.2495),
        (2439.4165,1620.25),
        (2439.4185,1620.2493),
        (2439.4197,1620.2494),
        (2439.4197,1620.2496),
        (2439.4192,1620.2496),
        (2439.4194,1620.2498),
        (2439.4192,1620.2499),
        (2439.4194,1620.2499),
        (2439.4192,1620.2501),
        (2439.419,1620.25),
[...]

PR #482, opened 2019-07-13
Add two failing tests

nical:
Test cases kindly provided in #481 and #476.


nical:
These two tests were in the new tessellator branch I just merged (they pass now!).

PR #495, opened 2019-11-17
Resolve merge events right of the current point during right events.

### LINKED ISSUE #492: Incorrect tessellation of path w/ holes
Tessellation of this shape with holes produces unexpected results.

This was the minimal example I was able to produce.  Gist, modified from glium basic example:
https://gist.github.com/Herschel/a7088696f8c89d831f07ce820c715e28

Current output (lyon 0.14.1):
image

Expected output-ish (Flash Player):
image

Seems like the tessellator adds a triangle jumping back to a previous subpath? I also tried adding variations of `PathEvent::Close` at the end of the subpaths, but no luck.

(cc https://github.com/ruffle-rs/ruffle/issues/77)

nical:
Fixes #492 

This was an oversight in the algorithm. Merge vertices must be resolved at the closest vertex that isn't separated fro the merge event by another edge, and that vertex can be on a right event which was not handled before.

### DIFF
--- a/tessellation/src/path_fill.rs
+++ b/tessellation/src/path_fill.rs
@@ -687,6 +687,8 @@ impl FillTessellator {
                     let edge_to = self.pending_edges[0].lower;
                     vertex_id = self.add_vertex_with_normal(&edge_to, &vertex_above, output)?;
                 }
+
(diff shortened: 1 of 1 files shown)


Herschel:
Thank you!

PR #498, opened 2019-12-13
Get rid of dependency on glsl-to-spriv

artursapek:
`glsl-to-spriv` is deprecated, and compiling these Spir-V files at runtime is not ideal. This PR commits pre-compiled `.spv` binaries, cleans up the directory structures, and updates the `wgpu` example code to just load these binary files directly.


nical:
Thanks!

PR #510, opened 2019-12-27
Replace dead url reference with title and citeseer url

jrmuizel:
http://cis.usouthal.edu/~hain/general/Publications/Bezier/Bezier%20Offset%20Curves.pdf is dead.

The version at http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.106.5344&rep=rep1&type=pdf is not the same as 
https://web.archive.org/web/20170911131001/http://cis.usouthal.edu/~hain/general/Publications/Bezier/Bezier%20Offset%20Curves.pdf but I assume it's better to link to the actual journal paper instead of what I assume is the preprint above.


nical:
Thanks!

PR #535, opened 2020-02-07
Use MSAA in wgpu example.

HalfVoxel:
Since for most people anti-aliasing of some form is a must when presenting something user facing it seems reasonable that at least one example should use it.

It adds a bit of code, but hopefully the feature makes up for it.


nical:
Thanks!

PR #544, opened 2020-02-16
Format code using 'cargo fmt'

Atul9:


nical:
+26,848 −4,873 is a little brutal but I suppose this will make running cargo format easier in the future.

PR #548, opened 2020-03-12
Implement Raph Levien's quadratic bézier flattening algorithm.

nical:
Replaces the quadratic bézier flattening implementation with @raphlinus's, described in https://raphlinus.github.io/graphics/curves/2019/12/23/flatten-quadbez.html 

It addresses an annoying issue with the previous implementation which was that the last segment was sometimes very small as a result of the iterative approach.

The only issue so far is that it does not handle the case with aligned control and endpoints where the control point is not between the endpoints (this currently defaults to a single line between the endpoints).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -432,7 +432,7 @@ name = "geom_bench"
 version = "0.0.1"
 dependencies = [
  "bencher 0.1.5 (registry+https://github.com/rust-lang/crates.io-index)",
- "lyon 0.15.5",
+ "lyon 0.15.6",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


raphlinus:
On a quick skim this looks good. You might want to take a look at line 270 of https://levien.com/tmp/flatten.html and see if that improves the colinear case.

Happy to see this implemented, and glad the writeup / code sample seemed to be adequate to explain the idea. The link above does more sophisticated things for cubics, fyi.

nical:
Thanks a lot @raphlinus!

I'll look into this code and maybe follow the approach of splitting cubic béziers into quadratic ones for flattening. Lyon's tessellator depends on being able to flatten and know the curve parameter of the original curve where flattening insert points, hopefully I can still extrapolate curve parameters of the cubic curve from the parameters in the quadratic pieces.

In the mean time I just noticed that lyon's current quadratic bézier flattening code also do not handle cusps so no need to hold off landing this.

nical:
```
test flatten_quadratic_levien ... bench:      11,233 ns/iter (+/- 1,051)
test flatten_quadratic_old    ... bench:      85,328 ns/iter (+/- 9,046)
```

Nice perf win!

PR #549, opened 2020-03-12
Fix #547

### LINKED ISSUE #547: Incorrect curve flattening
I found a case where the stroke tesselation is extremely incorrect.

What happens is that on a bezier curve it seems to only add a single vertex and then skip the remaining ones even though it is clearly not within tolerance.

```
let mut builder = Path::builder();
builder.move_to(point(9.8589325,53.186916));
builder.cubic_bezier_to(point(10.3262615,56.03796), point(8.514468,58.483364), point(7.0338364,60.40962));
builder.cubic_bezier_to(point(5.5532045,62.335873), point(6.1438327,61.547035), point(3.9364057,60.891937));
```

In the image below you can see the control points for a bezier curve and the incorrect tesselation.
It does not seem to depend on the tolerance given to the path tesselator. It is however extremely sensitive to the exact positions of the control points of the curve.

Screenshot from 2020-03-10 12-11-06

HalfVoxel:
This fixes #547 which was caused by a curve starting very close to an inflection point, or at least having a very small second derivative.

The included test checks if the flattened curve is within the given tolerance to the original curve. Before this fix it was at about 0.38 (the tolerance was 0.01) and after the fix it is at 0.016. So the test still fails unfortunately. According to the documentation the flattened curve should be at most a distance of #tolerance from the original curve, which it seems still does not hold. So I am leaving the test as it is.

Below is a screenshot of the test case before and after the bug was fixed. Note that there are 2 cubics in this picture. The second one starts at the right endpoint of the straight line (which is veeery close to being an inflection point). An inflection point is also indicated in magenta.
image
image

### DIFF
--- a/geom/src/flatten_cubic.rs
+++ b/geom/src/flatten_cubic.rs
@@ -179,9 +179,21 @@ fn flatten_including_inflection<S: Scalar, F: FnMut(Point<S>)>(
     tolerance: S,
     call_back: &mut F,
 ) -> CubicBezierSegment<S> {
-    let (before, mut after) = bezier.split(up_to_t);
+    let (mut before, mut after) = bezier.split(up_to_t);
+
+    // Approximate the start of #before with a straight line segment.
+    // This is necessary because at the start we *might* be close to an inflection point
+    // and this would cause #flatten_cubic_no_inflection to sometimes fail due to floating point errors.
+    if let Some(tf) = inflection_approximation_range(&before, tolerance) {
+        before = before.after_split(tf);
+        call_back(before.from);
+    }
+
     flatten_cubic_no_inflection(before, tolerance, call_back);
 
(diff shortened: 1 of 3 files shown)


HalfVoxel:
Actually... upon testing again it seems like while the specific test case in #547 is fixed by this, there are still cases where the flattening fails. I'll have to do some more tests...

nical:
Thanks a lot! I'll have a closer look shortly. There are a couple of test failures in tests that check that the flattening iterator and callback apis behave the same. It should be fixed by adding the same logic to the flattening iterator.

> there are still cases where the flattening fails

That's unfortunate. I'm considering switching to a different algorithm if I can find one that is more robust while not sacrificing too much speed and efficiency.

HalfVoxel:
This has been fixed by other commits. Closing.

PR #552, opened 2020-03-14
Implement Raph levien's cubic-to-quadratic bézier conversion.

nical:
Inspired from https://levien.com/tmp/flatten.html though it is missing part of the logic to deal with cusps.

The code is pretty simple as the number of quadratic bézier segments is to satifsfy the approximation is computed up-front and the curve is then subdivided at regular intervals.
It's 30% faster than than iteratively splitting ranges in half until the local approximation satisfies the tolerance threshold.

I have no idea why the math work, though. My intuition was that some parts of the cubic curves (cusp, inflections) would need finer subdivision than other areas but somehow the logic around line 256 of https://levien.com/tmp/flatten.html manages to get an upper bound on error accross the whole curve and with a gloriously low number of instructions.

@raphlinus do you plan to write a blog post about this as you did with the quadratic bézier flattening? It would be great to be able to reference the original explication in the code.

Out of curiosity, is the simplicity of this "global" subdivision scheme motivated by data parallelism and doing it on the GPU? Or does it also happen to be close to optimal as your quadratic flattening approximation?

### DIFF
--- a/geom/src/flatten_cubic.rs
+++ b/geom/src/flatten_cubic.rs
@@ -1,83 +1,88 @@
 ///! Utilities to flatten cubic bézier curve segments, implemented both with callback and
 ///! iterator based APIs.
-///!
-///! The algorithm implemented here is based on: "Fast, precise flattening of cubic Bézier path and offset curves"
-///! http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.106.5344&rep=rep1&type=pdf
-///! It produces a better approximations than the usual recursive subdivision approach (or
-///! in other words, it generates less points for a given tolerance threshold).
 
 use crate::generic_math::Point;
 use crate::scalar::Scalar;
 use crate::CubicBezierSegment;
+use crate::cubic_to_quadratic::single_curve_approximation;
+use crate::quadratic_bezier::FlattenedT as FlattenedQuadraticSegment;
 
-pub fn square_approximation_error<S: Scalar>(curve: &CubicBezierSegment<S>) -> S {
-    // See http://caffeineowl.com/graphics/2d/vectorial/cubic2quad01.html
-    S::THREE / S::value(1296.0)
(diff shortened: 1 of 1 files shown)


raphlinus:
I am actually planning on expanding it into a paper for the [jcgt](http://jcgt.org/), but this will take some time, and my bandwidth is pretty limited right now. I can appreciate how it must feel a bit like magic :)

> Out of curiosity, is the simplicity of this "global" subdivision scheme motivated by data parallelism and doing it on the GPU? Or does it also happen to be close to optimal as your quadratic flattening approximation?

I'm not sure I fully understand the question, but I think the answer is "both." It is absolutely motivated by data parallelism and GPU-friendliness, and it is also the case that it is nearly-optimal as is the quadratic version. For people following along at home, it's better than the version in the quadratic blog post because it also has improved special-case handling for cusps (it uses a more appropriate approximation when the segment includes a vertex of the parabola).

raphlinus:
Ah, the logic for computing the number of *quadratics* needed is nothing special, it's basically the same thing as the Sederberg and Wang references mentioned at the end of the quadratic blog post.

nical:
> I am actually planning on expanding it into a paper for the jcgt

Neat!

Thanks for the clarification. From some very unscientific testing, the difference between sampling from the quadratic and the cubic is quite small but it could be that I didn't test on extreme enough test cases. In fact, having both a callback and an iterator based API for flattening, I ended up sampling the quadratic in the former and the cubic in the latter out of convenience, but I might just sample the quadratic if only for consistency. 

> I'm not sure I fully understand the question, but I think the answer is "both." It is absolutely motivated by data parallelism and GPU-friendliness, and it is also the case that it is nearly-optimal as is the quadratic version.

Cool, my thinking was that if facing a tradeoff between data-parallelism and efficiency you had leaned towards data parallelism I would have probably spent some time experimenting to see if I could get more efficiency in my particular CPU-only case, but it seems that I won't need to do that which is great.

PR #571, opened 2020-04-24
Use Github Actions for CI

bnjbvr:
This is a proposal to use Github actions for CI, instead of Travis CI. I have found that Github Actions are faster to start, run, and that they're more reliable in general.

What this does:

- compile and run tests with each of Rust stable and nightly
- compile and run wasm tests with the Rust nightly toolchain

Here's an example of a successful run: https://github.com/bnjbvr/lyon/pull/2
And a failing run (contains a panic in a test case), showing the full backtrace on error: https://github.com/bnjbvr/lyon/pull/3

This is now ready for review. @nical PTAL!


nical:
Merci!

PR #588, opened 2020-06-07
Bump wgpu version to "0.5.0" in examples.

mr1sunshine:


mr1sunshine:
Hey, @nical 👋  May I ask you to review those changes? I was doing my own project and realized that it would be great to get lyon examples to work with latest `wgpu`. 
Although I have two problems at the moment:
1. Actually rendering picture is upside-down 😞  Couldn't find the reason yet. Any clue?
2. I'm not sure that the way I'm converting vertices for background geometry is optimal (`Vec<Point>` -> `&[u8]`). May I ask you to take a look?
Big thanks 🙏

nical:
Thanks! It's odd that the content is shown upside down. It doesn't look like a mistake in your code, I suppose the behavior changed in wgpu, they have to fiddle with different GPU APIs with different y-up/y-down convention so perhaps they changed their changed the convention at some point. That can be fixed up in the shader, I'll take your pull request and do the flip.

Thanks again!

PR #596, opened 2020-07-08
Fix abort condition for line segments in path walking

tronical:
When the callback returns that it is done, we should not continue
on the current line segment.


nical:
Nice, thanks! Please address the warnings in the test and then we're good to go.

tronical:
Apologies about the warnings - oops :).

nical:
No worries, thanks a lot!

PR #598, opened 2020-07-31
Updated to euclid 0.22

avr1254:
Updated to work with euclid 0.22. This is part of trying to update all servo dependencies to use euclid 0.22


nical:
Thanks! The master branch is in flux right now and not ready to be pucblished so what I'll do to unblock updating servo is also update euclid on top of the last published version without all of the newer stuff in master. I'll find some time for it this weeked.

PR #600, opened 2020-08-04
Use shields.io instead of meritbadge

nasso:
meritbadge.herokuapp.com doesn't use https and redirects to img.shields.io anyway. shields.io already offers support for Rust crate version fetching from crates.io: switching to shields.io gets rid of the "Connection not secure" warning that shows up on docs.rs in some browsers.


nical:
Alright, thanks!

PR #605, opened 2020-09-13
 Adding euclid_mint feature flag to enable euclid/mint

halzy:
Euclid has a feature called 'mint', this PR enables the feature flag piping to use that `euclid` feature.


nical:
Hi, thanks for this PR. Can't this be achieved outside of lyon by just having the crate that uses lyob add a dependency to euclid with the mint feature enabled?

halzy:
Oh, that's a great idea. I'll try it out.

halzy:
Ok, that works. Thank you! (closing PR)

Issue #611, opened 2020-12-12
Remove .clone() on a type that supports copy.


martinfrances107:
There is one clippy check -- that I repeatedly find leads to performance improvements.

It does not always follow .. it really depends on the situation.. but I find it useful enough that I always remove that clippy warning as standard.

There are 2 places where we call 

a.clone() 

when 'a' supports copy

I am about to push a PR.

As a general discussion point ..

I would be willing to put the time into clear all clippy warnings. Is that something the project as a whole would welcome?

PR #612, opened 2020-12-12
#611 Remove .clone() on a type that supports copy

martinfrances107:
This clears all  clippy warnings of this type in the project.


nical:
Thanks!

Issue #615, opened 2020-12-13
Clippy: Redundant field names


martinfrances107:
This is a minor code tidy 

```
let bar: u8 = 123;

struct Foo {
    bar: u8,
}

let foo = Foo { bar: bar };

```

bar:bar can be simplified. here is the clippy warning.

[redundant field names ](https://rust-lang.github.io/rust-clippy/master/index.html#redundant_field_names)

I am about to push a PR.

PR #616, opened 2020-12-13
#615 Removed redundant field names

martinfrances107:
As described in the issue 
#615 
`let foo = Foo { bar: bar };`
bar:bar is redundant.


nical:
Thanks!

PR #629, opened 2021-01-20
Soundness fixes

eduardosm:
* Correct lifetimes in iterator constructors.
* Use `wrapping_add` instead of `add` when the pointer can wrap (otherwise it is UB even if the result is never dereferenced).
* Avoid creating invalid slices (it is UB even if they are never dereferenced).

### DIFF
--- a/path/src/commands.rs
+++ b/path/src/commands.rs
@@ -89,7 +89,7 @@ struct CmdIter<'l> {
 }
 
 impl<'l> CmdIter<'l> {
-    fn new(slice: &[u32]) -> Self {
+    fn new(slice: &'l [u32]) -> Self {
         let ptr = slice.as_ptr();
         let end = unsafe { ptr.add(slice.len()) };
         CmdIter {
@@ -629,7 +629,7 @@ pub struct Iter<'l> {
 }
 
 impl<'l> Iter<'l> {
-    fn new(cmds: &[u32]) -> Self {
+    fn new(cmds: &'l [u32]) -> Self {
         Iter {
             cmds: CmdIter::new(cmds),
             idx: 0,
--- a/path/src/path.rs
+++ b/path/src/path.rs
@@ -748,7 +748,7 @@ struct PointIter<'l> {
 }
 
 impl<'l> PointIter<'l> {
-    fn new(slice: &[Point]) -> Self {
+    fn new(slice: &'l [Point]) -> Self {
         let ptr = slice.as_ptr();
         let end = unsafe { ptr.add(slice.len()) };
         PointIter {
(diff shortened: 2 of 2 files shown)


nical:
Thanks! I'm curious about the wrapping_add part. Was it detected by a tool? I don't think that it is possible to cause the overflow with the path API.

eduardosm:
> I don't think that it is possible to cause the overflow with the path API.

Neither do I. It is just to "be safe" by making sure the panic is raised before causing the undefined behavior.

nical:
To my knowledge, unlike in C, integer overlow is not undefined behavior in rust. I would prefer to keep the `add` call as it is, because `wrapping_add` makes it look like we expect it might overflow which is not the right signal to someone who is looking to understand the code. If overflow turned out to be possible through future mistakes in the code, `add` at least gives us a debug assertion.

eduardosm:
Pointer arithmetic is different from integer arithmetic. The [pointer::add](https://doc.rust-lang.org/std/primitive.pointer.html#method.add) function produces undefined behavior on overflow.

nical:
Sorry for the delay. Thanks for mentioning the pointer arithmetic difference, I didn't know that. I still prefer to keep the original `add` considering that the overflow cannot be crafted using the API. Otherwise in six months I'll be spending time wondering why I thought overflowing this pointer was a possible scenario (and whether I can benefit from more aggressive compiler optimizations via `add`). If some hardening can be done it should be to make extra sure the overflow never happens to avoid getting in into a state that is pretty broken even without UB.

eduardosm:
I have realized that my solution does not even fix the issue. If `n` is large enough, it might wrap around causing `ptr < self.ptr` and `ptr < self.end`, which won't be detected by the `assert`. I will push another commit that brings back `add` and properly checks for the remaining length.

eduardosm:
Regarding performance,
```rust
fn advance_n(&mut self, n: usize) {
    unsafe {
        assert!(self.remaining_len() >= n);
        self.ptr = self.ptr.add(n);
    }
}
```
produces
```
test simple_path_iter             ... bench:       7,552 ns/iter (+/- 160)
```

and
```rust
fn advance_n(&mut self, n: usize) {
    unsafe {
        self.ptr = self.ptr.add(n);
        assert!(self.ptr <= self.end)
    }
}
```
produces
```
test simple_path_iter             ... bench:      32,174 ns/iter (+/- 1,346)
```

nical:
I ran the benchmark and while I am not seeing a big difference as you do (both versions are within noise of each other on my laptop), it's not causing a regression either so I'm fine with taking this version of the PR.

Thanks!

PR #631, opened 2021-01-26
Fix relative quadratic bezier curve SVG commands

tronical:
Make sure to update the current position after the command so that
subsequent commands start at the right location.


tronical:
Thanks, that was quick :)

PR #640, opened 2021-03-10
Reducing the error of circles and rounded_rectangles

halzy:
This PR reduces the approximation error when drawing rounded rectangles or circles.

Uses the better approximation found here:
https://spencermortensen.com/articles/bezier-circle/

### DIFF
--- a/path/src/builder.rs
+++ b/path/src/builder.rs
@@ -1069,39 +1069,31 @@ fn add_circle<Builder: PathBuilder>(
         Winding::Negative => -1.0,
     };
 
-    let tan_pi_over_8 = 0.41421356237;
-    let cos_pi_over_4 = 0.70710678118;
-    let d = radius * tan_pi_over_8;
+    let constant = 0.551915024494;
+
+    let d = radius * constant;
 
     builder.begin(center + vector(-radius, 0.0));
 
     let ctrl_0 = center + vector(-radius, -d * dir);
-    let mid_0 = center + vector(-1.0, -dir) * radius * cos_pi_over_4;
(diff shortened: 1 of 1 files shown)


nical:
Hi, thanks for the PR!

My initial reaction is that the arc approximation using two quadratic curves is already quite good and super easy to reason about even though more involved approximations like the one you propose are more precise (the code is simple but the math behind it is not something I'd come up with easily). Have run into issues with the current level of precision?
Unless it is causing issues my preference would be to keep this code as simple as possible over getting the error from 0.03% to 0.02%.

Also, having the arc approximation slightly inflating the circle tends to balance the bézier flattening approximation which is biased the other way around (the error being always on the interior side of the curve). So unless the arc is very large the simple approximation probably (accidentally) gives a better approximation after flattening.

halzy:
Hi, Nical!

I'm building some shapes that end up inside a 3d space, so they can be quite large on screen. When closer to them the rounded corners didn't seem correct so I stacked a circle drawn in 'pie' segments on top of it.

The foreground object is a circle:
https://pasteboard.co/JS1OP55.png

vs the corrected version:
https://pasteboard.co/JS1Pwl0p.png

I was able to 'dial in' a closer approximation with the cubic bezier version by hand and just 'eyeing' it. I looked up the math after the fact. 

You have some good points. It's up to you if you want the changes. :-) 

Thank you for the terrific library!

nical:
Fair enough, it makes a bigger difference in a legit real-world use case than I anticipated, you convinced me. Thanks again!

Issue #36, opened 2017-05-21
Make the source code easier to unserstand and discover.


nical:
A followup issue from discussions in #32.

Relevant comment from @Drakulix:
> [...] I have a hard time to figure out what parts of it can be used in combination and how they should fit together.
> 
> [...] I don't feel like I would know where to start without studying the examples in great detail at first (which was even more difficult, before you added the new simpler example).
> 
>  This could most likely be resolved by some explanation and maybe some smaller examples in the giant meta-crate. That would be the place, I would expect a gentle introduction about how to start using this project.

Looks like we need at least:
 - [ ] A "getting started" section in the documentation of the meta-crate with examples.
 - [ ] More standalone examples preferably as simple and focused as possible.
 - [x] More detailed explanations about which crates are useful for what, which crates are optional and how the crates interact with one another.

Drakulix:
This has improved so much with the last release. Thanks! :)

nical:
You're welcome and thank you too! These improvements were made thanks to the feedback you gave. The docs are still far from perfect so don't hesitate to point me to any of its shortcomings.

LU15W1R7H:
I think the codebase would also benefit from some restructuring. This would make the "discovering" easier.
Some things are not as organized, which makes reading the code harder.

Issue #32, opened 2017-05-21
Make the project more welcoming and appealing to contributors and users.


nical:
Some rust projects have done a stellar job at moving from being a one-person project into being owned and contributed to by a small community built around it. The [gfx crate](https://github.com/gfx-rs/) is a prime example of this.

I would very much like lyon to be more community-driven, so that the project advances faster and in a direction that is useful to more people, and have more people to brainstorm ideas with.

Some ideas off the top of my head:
 - Good documentation: a never ending endeavor, but there has been a lot of work in this area lately.
 - Make lyon easier to use: The tools provided like tessellators are (purposefully) quite low-level and it is a lot of work to use them in any project. Building higher level APIs on top of this could maybe help with being useful for more people? I am experimenting with higher level APIs currently (issue #17).
 - File more issues for newcomers to pick, advertise them in this-week-in-rust.
 - Make *better* issues (more detailed) to help contributors getting started.
 - Implement more functionalities: There is still a ton of things to do, perhaps some people are waiting for some key missing features to be implemented?
 - Make more interesting demos that show what lyon can offer.
 - Move the repository from my github account into a dedicated github organization.
 - Reduce the complexity of the code if possible.

Drakulix:
> File a more issues for newcomers to pick, advertise them in this-week-in-rust.

This. I would like to contribute sometime (although I have too many projects in general), but I have no idea, where to start, as the code base appears large and overwhelming.

More comments on the "easy" issues on how to approach it, would be very helpful. E.g. where to add the new functionality (files), how should the api look like, some implementation tipps for the underlying project.

> Implement more functionalities: There is still a ton of things to do, perhaps some people are waiting for some key missing features to be implemented?

I was mainly waiting for all the basic shapes to be fully implemented (especially rounded rectangle) as well as round line endings.
After that I wanted to prototype a game I wanted to build for a long time with this crate, that is the biggest reason I am watching this repository.

nical:
Thanks a lot for your feedback! I'll start with adding details to the the issues labelled as easy and spend some time on the basic shapes, then.

Are there specific obstacles that make the code base appear overwhelming? Is it hard to understand the scope/goal of the project, the structure of the code and the repository, finding useful information in the documentation, or understanding the algorithms?

Drakulix:
Mostly structure. It is nice, that the project is this modularized, but I have a hard time to figure out what parts of it can be used in combination and how they should fit together.

[...]

PR #645, opened 2021-04-15
Restructuring repository

LU15W1R7H:
This PR restructures the repository.

All crates have been moved into a `crates/` directory.
This structure mirrors the layout of the [ripgrep](https://github.com/BurntSushi/ripgrep) repository.

I consider this to be part of the effort around issues #36 and #32.
It makes the repository more approchable since, the source code is neatly grouped together.
You can tell from once glance where all the source code is located.


LU15W1R7H:
What do you think? Is this controversial?

nical:
I'm a sceptical that it will make the code easier to understand, but it'll probably look more organized this way if the number of crates in lyon grows, so why not.

A few things, though:

- `cli` isn't a published crate like most things in the repo, it's a command-line app to test and debug things, so it should stay at the root of the repository.
- let's give the 'lyon` crate the usual structure with a `src/` folder. For example `crates/lyon/src/lib.rs` instead of `crates/lyon/lib.rs`.

LU15W1R7H:
Sounds good 👍🏾

PR #655, opened 2021-05-24
Fix example msaa argument

foobar27:
The msaa argument was effectively ignored, because of a mixup of min and max to add a lower bound of 1.


nical:
Woops! thanks!

PR #656, opened 2021-06-02
Fix SVG arc_to command not updating the current position

tronical:
As per spec, the to argument becomes the current point for subsequent
commands. This is mostly covered inside the builder, but for example the
relative_* functions on the SvgBuilder use the current_position.


nical:
Thanks !

The problem is also present in `arc` that `arc_to` uses under the hood, so I think it should be fixed there instead. One way to do this is to set the current position in the callback after the `quadratic_bezier_to` call.

tronical:
Hm, I was contemplating the same but wasn't sure since the `arc` API doesn't look like it would change the current position (there's no to parameter). But since you're pointing it out it makes sense then :-). I'll force-push that version shortly.

nical:
Thanks! I'll investigate #657 this weekend and publish a new version with both fixes soon.

PR #663, opened 2021-06-10
Bump `arrayvec` to 0.7

vihdzp:
This pull request bumps the `arrayvec` crate from version 0.5 to version 0.7. The only noticeable change is that `ArrayVec<[N; C]>` becomes `ArrayVec<N, C>`.


nical:
Hi, thanks! This is technically a breaking change. I try to group those to avoid frequent API breaks in lyon as much as possible, so I'll wait a little while before merging it. The new arrayvec ergonomics look nice!

dhardy:
Maybe it would be time to land this, plus a few more version bumps, soon? There are quite a few things which could be updated, most notably `wgpu`, but also e.g. `rand` (version 0.6 is quite old). I was considering opening a PR, but looks like this should land first...

nical:
`rand` and `wgpu` aren't part of public APIs, they are only used by the cli app and examples so we can update them at will whenever is convenient. I would prefer to hold off updating arrayvec for a little while longer and batch it with other upcoming breaking changes.

nical:
For some reason the github UI got buggy and I ended up having to manually merge this PR from git, so I'm closing the PR but the commit made it to the master branch: https://github.com/nical/lyon/commit/67ad0177d1efddfb4f5e7d6c451a9202a7ff2158

PR #621, opened 2021-01-06
Store Span's tesselator behind a pointer

adrian17:
`MonotoneTessellator` is a huge struct, so copying it around when pushing/popping and especially in `spans.retain()` is relatively expensive. Boxing it adds a layer of indirection and extra `Option` checks in other parts of the algorithm, but significantly speeds up `cleanup_spans`, `begin_span` and `end_span`.

Tessellation benchmarks see a 15-30% improvement. I can also see a similar 10-20% improvement in real world uses in https://github.com/ruffle-rs/ruffle in wasm builds.

Benchmarks before vs after:
```
fill_events_01_logo                 72,443 ns/iter (+/- 8,185)  ->  72,465 ns/iter (+/- 4,044)
fill_events_02_logo_pre_flattened   40,760 ns/iter (+/- 8,680)  ->  39,751 ns/iter (+/- 2,796)
fill_events_03_logo_with_tess      251,327 ns/iter (+/- 6,709)  -> 209,559 ns/iter (+/- 51,154)
fill_tess_01_logo                  213,336 ns/iter (+/- 9,482)  -> 168,523 ns/iter (+/- 5,023)
fill_tess_03_logo_no_intersections 230,245 ns/iter (+/- 7,912)  -> 185,115 ns/iter (+/- 11,044)
fill_tess_05_logo_no_curve         132,274 ns/iter (+/- 6,337)  ->  92,060 ns/iter (+/- 13,499)
fill_tess_06_logo_with_ids         210,871 ns/iter (+/- 27,150) -> 168,993 ns/iter (+/- 11,192)
flattening_01_logo_simple_iter      41,266 ns/iter (+/- 923)    ->  40,935 ns/iter (+/- 2,431)
flattening_02_logo_iter             51,051 ns/iter (+/- 2,705)  ->  52,444 ns/iter (+/- 2,448)
flattening_03_logo_builder          39,201 ns/iter (+/- 2,078)  ->  39,273 ns/iter (+/- 1,429)
stroke_01_logo_miter                67,696 ns/iter (+/- 3,201)  ->  67,516 ns/iter (+/- 4,866)
stroke_02_logo_bevel                72,127 ns/iter (+/- 3,793)  ->  73,016 ns/iter (+/- 4,191)
stroke_03_logo_round                81,014 ns/iter (+/- 11,486) ->  81,680 ns/iter (+/- 6,663)
```

I tried running tests, but `cargo test` only saw 4 doctests, and ignored 2 of them - is this right?

[...]

PR #664, opened 2021-06-16
Got rid of clippy errors

OfficialURL:
This PR gets rid of all warnings and errors thrown by `cargo clippy`. Most of what this tool does is standardize a Rust style of code, though it also catches some trivial optimizations. I'll comment on a few recurring categories of suggested lints:

- There were quite a few warnings about using `return` at the end of a method. This is ultimately a matter of style, but in general, removing noise will make your code a tiny bit easier to read for others.
- There were also a few instances of clippy complaining about `&slice[..]` where `slice` would have sufficed. I don't really know why you chose the former notation.
- Calling `option.unwrap_or(arg)` eagerly evaluates `arg`, even if it's called on a `Some`. If `arg` has non-zero cost, it's always better to use `option.unwrap_or_else(|| arg)` instead.
- A few methods in the code that had `new` methods with no arguments, but didn't implement `Default`. In general, one should implement `Default` whenever reasonable, as it allows users of the crate to use `#[derive(Default)]` in more cases.
- Some structs had a `_private` field, meant so that adding fields to the struct wouldn't be a breaking change. However, adding `#[non_exhaustive]` at the struct declaration is simpler and achieves the same effect.
- There were some implementations of the `Into` trait, which doesn't automatically implement the corresponding `From` trait. However, implementing `From` automatically implements the corresponding `Into` trait. 
- There were some C-style iterations over vectors where iterators sufficed. Sometimes the iterators made the code quite simpler, though some of these are kinda dubious.
- There were a lot of struct field instantiations of the form `x: x`, when Rust allows you to just write `x` instead for clarity. I think this one is an absolute win.
- Clippy was complaining about comparing floating points strictly, but since the code seems to depend on this a lot, I disabled this error.

Some specific warnings/errors which I still think are noteworthy:

- The methods in the tessellation code returned types like `Result<T, ()>`. Unwrapping these calls `dbg!(())` in case of failing, which doesn't really say anything, and displaying the error is equally unhelpful. I instead made a unit `TessellationError` struct, which displays `an unknown error occurred during tessellation` if printed.
- `MonotoneTessellator` was sometimes `Box`ed, even though it was already on the heap. I don't really know what the justification for this was. Perhaps size? Feel free to revert this one.

And some other fixes that weren't suggested by `clippy`, but which I still think make the code clearer.

- At times you were taking redundant references to `&str`. I removed these for clarity.
- Some methods in `AdvancedPath` took `&mut AdvancedPath` as an argument, but using `&mut self` achieves the same effect.

I also wanted to put the various magic constants throughout the code in `const` declarations, but honestly, I'm not sure what most of them do.

I'll leave to your discretion which of these changes you're willing to implement. I'd advise that you add `#![allow(lint)]` on the crates' root files for all lints you want to ignore.


crates/geom/src/arc.rs
```diff
@@ -1049,20 +1049,4 @@ fn negative_flattening_step() {
     };
 
     arc.for_each_flattened(0.100000001, &mut |_| {});
-
-    // There was also an issue with negative sweep_angle leading to a negative step
```
nical, line 1053: It looks like you removed this part of the test. Was it a rebasing error?

crates/path/src/builder.rs
```diff
@@ -89,6 +89,9 @@ use crate::{EndpointId, Winding};
 use std::iter::IntoIterator;
 use std::marker::Sized;
 
+/// https://spencermortensen.com/articles/bezier-circle/
+const BEZIER: f32 = 0.55191505;
```
nical, line 93: This is quite specific to the logic of the bezier approximation so I would rather keep this near the code that uses it otherwise it is hard to understand what it is about (especially with a name like BEZIER). You could roll it back to how it was (declared twice), but using a `const` instead of a regular immutable variable.

crates/path/src/events.rs
```diff
@@ -39,13 +39,13 @@ pub type IdEvent = Event<EndpointId, ControlPointId>;
...
-        match self {
+        matches!(
+            self,
             &Event::Line { .. }
-            | &Event::Quadratic { .. }
-            | &Event::Cubic { .. }
-            | &Event::End { close: true, .. } => true,
-            _ => false,
-        }
+                | &Event::Quadratic { .. }
```
nical, line 45: I'm not very fond of this extra indentation.

crates/tess2/src/tessellator.rs
```diff
@@ -11,6 +11,18 @@ use std::ptr;
 use std::slice;
 use tess2_sys::*;
 
+/// A unit struct that represents any error occuring during tesselation.
+#[derive(Debug)]
+pub struct TesselationError;
```
nical, line 16: In principle I think this is a good idea but I want to delay API breaking changes until a while later, So it would be great to remove them or put them out in another PR to avoid delaying everything else.

crates/tessellation/src/fill.rs
```diff
@@ -174,11 +174,11 @@ struct ActiveEdges {
 }
 
 struct Span {
-    tess: Option<Box<MonotoneTessellator>>,
+    tess: Option<MonotoneTessellator>,
```
nical, line 177: This is actually undoing a performance optimization (see #621), let's roll it back.

crates/tessellation/src/fill.rs
```diff
@@ -2060,7 +2063,7 @@ pub(crate) fn compare_positions(a: Point, b: Point) -> Ordering {
     if a.x < b.x {
         return Ordering::Less;
     }
-    return Ordering::Equal;
+    Ordering::Equal
```
nical, line 2066: Blank line before return expression.

crates/tessellation/src/fill.rs
```diff
@@ -2206,7 +2207,7 @@ impl<'l> FillVertex<'l> {
         }
 
         if div > 1.0 {
-            for attribute in &mut self.attrib_buffer[..] {
+            for attribute in self.attrib_buffer.iter_mut() {
```
nical, line 2210: It's nice when the IntoIter stuff figures out that [..] can be removed but when it can't I prefer `&mut foo[..]` over `foo.iter_mut()`.

crates/tessellation/src/geometry_builder.rs
```diff
@@ -524,19 +529,21 @@ impl GeometryBuilder for NoOutput {
 
 impl FillGeometryBuilder for NoOutput {
     fn add_fill_vertex(&mut self, _vertex: FillVertex) -> Result<VertexId, GeometryBuilderError> {
-        if self.count.vertices >= std::u32::MAX {
+        // This code is never ran!
```
nical, line 532: Do you mean that the case where count.vertices == std::u32::MAX is unreachable ?

crates/geom/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "lyon_geom"
-version = "0.17.2"
+version = "0.17.1"
```
nical, line 3: This looks like a rebasing error.

nical (changes requested):
Thanks a lot! There's a lot of good stuff in there.

Some of this code dates from rust's earlier days and some parts are showing their age for sure.

There's a few things to change, some sytlistic and others a bit more important. I noticed that at least one of the improvements introduces an API breaking change, I haven't done a very thorough pass on those but if you notice others please extract them out of the patch, I make and effort to avoid frequent breaking crate version bumps.

cli/src/fuzzing.rs
```diff
@@ -43,7 +42,7 @@ fn generate_path(cmd: &FuzzCmd, iteration: u64) -> Path {
             break;
         }
     }
-    return path.build();
+    path.build()
```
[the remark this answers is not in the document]
OfficialURL: Hehe don't worry, I do the same on my own code. I just didn't add the spacing just in case.

crates/tessellation/src/geometry_builder.rs
OfficialURL: I misunderstood Clippy's suggestion, actually. It's rather the case where `count.vertices > std::u32::MAX` that's unreachable. I'll uncomment the code and change it to an equality.

crates/path/src/events.rs
OfficialURL: I can't really get rid of the indentation unless we get rid of `cargo fmt` altogether, which I'd advise against. I could turn this back into a `match` statement and add an `#[allow(clippy::match_like_matches_macro)]` attribute, though.

crates/tessellation/src/fill.rs
OfficialURL: Clippy complains if I change it back to the former. It (incorectly) tells me I'm slicing redundantly. Should I disable the lint, or do we go for `iter_mut` in this particular instance instead?

crates/geom/src/arc.rs
OfficialURL: Yep, I've added it back.

crates/tess2/src/tessellator.rs
OfficialURL: I've made PR #665 with this change, and removed it from the current one.

crates/tessellation/src/fill.rs
OfficialURL: I've rolled it back, and added a doc comment referencing this issue so that it doesn't happen again.

crates/geom/Cargo.toml
OfficialURL: Yep, fixed.

crates/tessellation/src/fill.rs
nical: Bah, Ok let's not swim against the current. you can keep it as it is in your patch.

OfficialURL:
It seems like there's some newer lints in the latest version of Rust, but some of them will be breaking as well. I'll put their fixes in different PRs.

crates/extra/src/debugging.rs
```diff
@@ -34,19 +34,20 @@ pub fn path_to_polygons(path: PathSlice) -> Polygons {
             }
         }
     }
-    return polygons;
+    polygons
 }
 
-pub fn polygons_to_path(polygons: &Polygons) -> Path {
+pub fn polygons_to_path(polygons: &[Vec<Point>]) -> Path {
```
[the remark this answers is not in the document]
OfficialURL: I've moved this over to PR #667.

OfficialURL:
If this PR, along with #665, #667, and #668, get accepted, there will no longer be any `clippy` errors. 😄

nical:
This is looking good! One last thing: please squash this PR into a single commit and I'll merge it.

OfficialURL:
> This is looking good! One last thing: please squash this PR into a single commit and I'll merge it.

Everything has been merged into a single commit!

crates/geom/src/quadratic_bezier.rs
```diff
@@ -521,7 +521,7 @@ impl<S: Scalar> QuadraticBezierSegment<S> {
     /// The result is provided in the form of the `t` parameters of each
     /// point along curve. To get the intersection points, sample the curve
     /// at the corresponding values.
-    pub fn line_intersections_t(&self, line: &Line<S>) -> ArrayVec<[S; 2]> {
+    pub fn line_intersections_t(&self, line: &Line<S>) -> ArrayVec<S, 2> {
```
nical, line 524: I hadn't noticed before but the ArrayVec change has snuck in.

crates/tessellation/src/fill.rs
```diff
@@ -2183,8 +2191,8 @@ impl<'l> FillVertex<'l> {
                     let a = store.get(id);
                     assert!(a.len() == num_attributes);
                     assert!(self.attrib_buffer.len() == num_attributes);
-                    for i in 0..num_attributes {
-                        self.attrib_buffer[i] += a[i];
+                    for (i, &att) in a.iter().enumerate().take(num_attributes) {
```
nical, line 2194: This one is less readable. I'd prefer to `allow(clippy::needless_range_loop)` at the crate level since clippy isn't able to tell when the index is needed anyway.

nical (changes requested):
Sorry, hopefully the last round of chnages before this can be merged.

crates/tessellation/src/fill.rs
OfficialURL: I just realized that `.take(num_attributes)` was redundant. Do you still stand by this?

nical: It's still meh, but alright nothing is set in stone anyway, I can always change it later if it rubs me the wrong way.

crates/tessellation/src/fill.rs
```diff
@@ -192,19 +194,20 @@ impl Span {
 
 struct Spans {
     spans: Vec<Span>,
-    pool: Vec<Box<MonotoneTessellator>>,
+    pool: Vec<MonotoneTessellator>,
```
nical, line 197: This needs to be reverted to the Boxed version (the whole pull #621 optimization thing).

crates/tessellation/src/fill.rs
```diff
@@ -192,19 +194,20 @@ impl Span {
...
 }
 
 impl Spans {
     fn begin_span(&mut self, span_idx: SpanIdx, position: &Point, vertex: VertexId) {
-        let mut tess = self
-            .pool
-            .pop()
-            .unwrap_or_else(|| Box::new(MonotoneTessellator::new()));
+        let mut tess = self.pool.pop().unwrap_or_else(MonotoneTessellator::new);
```
nical, line 202: Same here.

nical (changes requested):
Sorry, I feel a little bad about how this PR is taking ages to be merged. Big diffs are hard to review properly.

I just noticed that the `Box<MonotoneTessellator>` wasn't fully reverted (it's not only the option thing, the vec is important as well). After that I think that it should be ready to land once you've squashed it into a single commit.

OfficialURL:
> Sorry, I feel a little bad about how this PR is taking ages to be merged. Big diffs are hard to review properly.

Thanks for taking your time. Sorry for making such abrupt changes all in one. I'll know to split things up next time.

OfficialURL:
Ugh, the arrayvec commit somehow keeps creeping into this. Let me fix this real quick.

nical (approved):

nical:
Looks like you need to rebase this PR on top of the master branch.

OfficialURL:
Rebase successful!

nical (approved):
Thank you for you patience!

PR #665, opened 2021-06-16
Added `TessellationError`

vihdzp:
As discussed in #664:

> The methods in the tessellation code returned types like `Result<T, ()>`. Unwrapping these calls `dbg!(())` in case of failing, which doesn't really say anything, and displaying the error is equally unhelpful. I instead made a unit `TessellationError` struct, which displays `an unknown error occurred during tessellation` if printed.

This would be a breaking change, but I also believe that it would be a nice enhancement, and might help catch future bugs more easily.

### DIFF
--- a/crates/geom/src/arc.rs
+++ b/crates/geom/src/arc.rs
@@ -1065,4 +1065,3 @@ fn negative_flattening_step() {
 
     assert!(flattened.len() > 1);
 }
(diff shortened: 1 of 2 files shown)


nical:
Sorry for the wait, let's start merging the breaking changes!

PR #666, opened 2021-06-17
Fix lyon_svg::path_utils::build_path not returning Err on invalid com…

tronical:
…mands

It's good behavior to let the caller know that the path building failed.

This helps with sixtyfpsui/sixtyfps#262


nical:
Woops! thanks!

tronical:
Would it be possible to also get a new release of lyon_svg on crates.io? Thanks :)

nical:
I had totally forgotten about this one, sorry. Thanks for the reminder! published in `lyon_svg 0.17.2`.

Issue #85, opened 2017-06-24
API guidelines: ad-hoc conversions follow as_, to_, into_ conventions.


nical:
This is a sub-issue for #44.

We should audit the docs/code of each lyon crate and check that [ad-hoc conversions follow `as_`, `to_`, `into_` conventions](https://github.com/brson/rust-api-guidelines#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv).
If some of the code happens to not follow the guidelines we should fix it.

PR #668, opened 2021-06-17
`&self` to `self` in methods of the form `to_X` in `Copy` structs

vihdzp:
This is related to issue #85. This would be a breaking change, though in practice there would rarely be any noticeable difference.

The `to_arc` method in `SvgArc` is notable, since that one then goes on to call `from_svg_arc`, which also takes a reference. Perhaps we may want to make the latter take the value itself as well?

### DIFF
--- a/crates/geom/src/arc.rs
+++ b/crates/geom/src/arc.rs
@@ -136,7 +136,7 @@ impl<S: Scalar> Arc<S> {
     }
 
     /// Convert to the SVG arc notation.
-    pub fn to_svg_arc(&self) -> SvgArc<S> {
+    pub fn to_svg_arc(self) -> SvgArc<S> {
         let from = self.sample(S::ZERO);
         let to = self.sample(S::ONE);
         let flags = ArcFlags {
@@ -487,8 +487,8 @@ impl<S: Scalar> Into<Arc<S>> for SvgArc<S> {
 
 impl<S: Scalar> SvgArc<S> {
     /// Converts this arc from endpoints to center notation.
-    pub fn to_arc(&self) -> Arc<S> {
-        Arc::from_svg_arc(self)
+    pub fn to_arc(self) -> Arc<S> {
+        Arc::from_svg_arc(&self)
     }
 
     /// Per SVG spec, this arc should be rendered as a line_to segment.
@@ -1065,4 +1065,3 @@ fn negative_flattening_step() {
 
     assert!(flattened.len() > 1);
 }
-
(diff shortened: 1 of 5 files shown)


nical (changes requested):
There is some nuance on the question of whether things should be passed by value beyond whether we can copy them "trivially" (at least at the syntactical level).

Rust kind of likes to pretend that moving/copying large-ish structs is free but in practice the copies/moves aren't. They have a small cost spread everywhere which adds up and show up in performance profiles. In the case of function argument, the calling convention imposes copying these large structs on the stack whereas when passed by reference the pointer is simply placed into a register (up to a certain number of arguments).

So long story short, after having spent a lot of effort tracking these things down in Firefox, my rule of thumb is more or less: "if it's 128bits or more, pass by reference". It saves me having to think about whether a function will be called in performance critical code and if clippy disagrees, then it shall be silenced.

So for most of the types in this PR I prefer to continue passing the structures by ref.

For the ID types it makes sense to pass them by value but it's technically an API breaking change. A small one that's unlikely to break much code in practice but since it's not fixing a real issue, I'd rather be principled and either put it in the "for a future API bump" bucket or just forget about it.

So in the short term I propose to just silence the lint at the crate level so that the master branch can be clippy-warning-free.

OfficialURL:
If these copies are expensive enough, it might be worth not even implement `Copy` in the first place. That way, we can be more conscious about their cost, and as a bonus, we'll be following the Rust API guidelines when using `&self`.

nical:
The compiler does it's best to ellide copies within function bodies as much as it can, but public methods have to abide to standard calling conventions so that code can interoperate outside of its crate. By definition there is no optimization the compiler can do to avoid copying large struct arguments on the stack unless the function is inlined. Unfortunately rust is particularly shy about inlining across crates.
There's a distinction in my humble opinion between the cost of copy assignments which the programmers write explicitly and can minimize by rearranging the code, and the cost of passing large arguments by value which happens implicitly and can't be worked around without forcing inlining.

Some of these types would be very bothersome to have to clone every time and being 100% compliant with this guideline isn't worth the breakage that it would cause, especially for something that is almost syntactically identical for users of the crate (passing self by value instead of ref).

So my take here is that guidelines aren't rules. It's fine to bend them when there are practical reasons, especially when it doesn't really affect users of the API (syntactically). Having already done the thorough profiling for this type of thing I feel pretty good about the reasoning here.
Another indication of the grey area around this particular guideline is that you can [configure a size threshold in clippy](https://rust-lang.github.io/rust-clippy/master/#trivially_copy_pass_by_ref) to decide what is a good candidate for passing by ref or value (I don't know if it affect the `as_`/`to_`/`into_` methods).

nical:
Cherry-picked parts of this in #709 (for the id types). The larger types will remained passed by reference.

PR #669, opened 2021-06-21
Fix panic when feeding SVG false arc command

tronical:
When loading an SVG with a path command where an arc has its destination
at the current position and the radii are very small, the SVG spec says
to render it as a line_to segment. This patch calls line_to in that
case, instead of causing a run-time panic.

This helps with sixtyfpsui/sixtyfps#266

### DIFF
--- a/crates/path/src/builder.rs
+++ b/crates/path/src/builder.rs
@@ -1059,7 +1059,7 @@ impl<Builder: PathBuilder> SvgPathBuilder for WithSvg<Builder> {
     }
 
     fn arc_to(&mut self, radii: Vector, x_rotation: Angle, flags: ArcFlags, to: Point) {
-        let arc = SvgArc {
+        let svg_arc = SvgArc {
             from: self.current_position,
             to,
             radii,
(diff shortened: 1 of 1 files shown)


nical:
Thanks!

tronical:
Thanks for the quick review & merge. Just curious: what git branch/tag do you make maintenance releases off of?

Asking because I’d like to crates-patch a git sha1 into our repo (until the next release).

nical:
I don't tag releases on github. Work gets merged in the master branch and if the branch contains unreleased API breaking changes I usually have another branch named after the previous release  where I backport and publish fixes (that's rather rare).

Usually I publish bug fixes on crates.io immediately after merging them. I was busy yesterday unfortunately so that got delayed but I just published `lyon_path 0.17.4`.
When these delays happen don't hesitate to ping me. I'd rather make sure people can work entirely with crates.io unless they want to try out an experimental rewrite that isn't ready yet or things like that.

tronical:
Wonderful. Thanks for the explanation and the frequent releases :)

PR #683, opened 2021-08-02
Make reset public so users can retain the fill tesselator

yamadapc:
Exposes `FillTesselator::reset` so the instance can be recycled between uses.


nical:
Hi, thanks for this PR.

No need to call reset manually, it's always called at the beginning of tessellation. Reusing the tessellator instance is indeed the recommended way since it avoids a lot of memory allocations on subsequent uses.

yamadapc:
I see. Thank you! Let me close this then & try without calling reset.

PR #684, opened 2021-08-03
Add further iterator support for paths

notgull:
This PR does the following:

- Implements `FromIterator` for `Path`, allowing it to be built from an iterator over `PathEvent`s without having to go through `Builder`.
- Adds a new `Iter` struct that iterates over the individual `PathSlice`s in a `PathBuffer`.
- Also implements `FromIterator` for `PathBuffer`.
- Adds the `size_hint` method to the `Flattened` iterator.

### DIFF
--- a/crates/path/src/iterator.rs
+++ b/crates/path/src/iterator.rs
@@ -201,6 +201,19 @@ where
             None => None,
         }
     }
+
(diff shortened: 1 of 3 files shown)


nical:
Nice additions. Thanks!

PR #693, opened 2021-08-19
Arc::to_svg_arc: fix swapped ArcFlags

sameer:
Based on https://www.w3.org/TR/SVG/paths.html#PathDataEllipticalArcCommands

large arc = sweep angle >= 180
sweep = sweep angle.is_positive()

But it looks like those two are flipped in the code. Running into some issues with curve splitting and flattening that works fine if I flip them back.


nical:
Indeed, thanks!

PR #696, opened 2021-08-29
Reintroduce some basic shapes fast paths

nical:
These use the old custom routines with the same `FillVertex` API as the fill tessellator. The vertex structure points to an empty event queue which doesn't need allocation and should gracefully fall back to providing nothing when querying vertex sources.

The main downside I see so far is that this conflicts with the way I hoped to partially support computing normals in #633 so most likely either fast basic shapes or vertex normals will make it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1225,7 +1225,7 @@ dependencies = [
 
 [[package]]
 name = "lyon_tessellation"
-version = "0.17.8"
+version = "0.17.9"
 dependencies = [
  "arrayvec",
  "float_next_after",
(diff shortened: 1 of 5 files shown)


hecrj:
Thanks!

Just tried this branch in https://github.com/hecrj/iced/pull/683 and the performance is very similar to `0.16` :tada: 

In any case, the generic tessellator approach isn't that slow. At least, not slow enough to be the bottleneck in the `iced` examples. Thus, I'd understand if these changes don't make it!

nical:
Good to know.

Thankfully this change doesn't modify the API and removing it wouldn't either so I'll merge it for now even if it might be removed in the future if it conflicts with important stuff.

PR #702, opened 2021-09-13
Add missing `serde::` prefix when enabling serialization in a few crates

bnjbvr:
If you tried to compile with `cargo build --all-features`, this would fail because of the unprefixed `Serialize`/`Deserialize` derive macros.


nical:
Merci!

PR #703, opened 2021-09-13
Remove unused dependencies

bnjbvr:
This doesn't lower the size of the Cargo.lock file, but it may change
the crate compilation order, and give cargo more opportunities for
compilation parallelism.


nical:
Merci!

PR #784, opened 2022-04-29
Fix backward advancement interpolation near the start of closed strokes

nical:


FuriouZz:
Thanks, I will have a look

nical:
Nice catch, thanks!

PR #786, opened 2022-06-07
Fix clipped_x check on line past range end

tarkah:
`start` was accidentally typed instead of `end` when checking if a line was past the end of the clip range. This caused undesirable clipping behavior when using this method.


nical:
Ouch! Thanks!

Issue #787, opened 2022-06-18
Should `PathWalker` be replaced?


Mivik:
I'm creating an animation of a sprite moving along a path, and that's when I found PathWalker impractical. If the moving process is linear then we can still use it, but in fact the moving process is non-linear (see examples from [https://easings.net/](https://easings.net/)) and thus simply tweaking `interval` won't yield useful results.

I wonder if we can adopt `PathMeasure`, the approach in Skia. And if not, why is that? Is the use of `PathWalker` required by some design principles or something else?

New to these things and sorry for my probably incorrect English.

nical:
Hi, are you working on top of the current crates.io release or the master branch? The path walker should let you move linearly (as in, at constant speed) along the path. However it is based on an approximation which you can tweak via a tolerance parameter. The lower the tolerance parameter, the better the approximation, at a performance cost. 

Try lowering the tolerance parameter (to some smaller value superior to 0.0) and let me know if it helps.

I just had a look at skia's `PathMeasure` code and it is very similar to the approach in lyon_algorithms. The most notable difference being whether or not to cache information for future use, and the fact that skia uses a fixed tolerance threshold which I suppose is intended to give acceptable precision when 1 unit of length corresponds to roughly one pixel.

Mivik:
I'm working on top of the current crates.io release. And I think you've probably misunderstood my issue.

So basically, I understand how `tolerance` works. It works as a threshold, lowering which would produce results with higher precision as the path is tessellated into more segments.

My problem is, it's hard to obtain information about **specific points** on a path using `PathWalker`. For example, if I want to know the "mid-point" of a path, I'll have to write:

```rust
let length = approximate_length(&path, 0.01);
let counter = 0;
let mut pattern = RegularPattern {
    interval: length * 0.5,
    callback: |pos, _, _| {
        counter += 1;
        if counter == 2 {
            println!("The mid-point is {:?}", pos);
        }
        true
    },
};
walk_along_path(path.iter(), 0.0, pattern);
```

And what I expect is something like:

```rust
let length = approximate_length(&path, 0.01);
let measure = path.measure();
println!("The mid-point is {:?}", measure.position_at(0.5));
```

This is crucial when implementing path animation. Imagine an animation of a point moving along a path. If `PathMeasure` is introduced, I can simply go like:

```rust
fn on_update(&self, progress: f64 /*from 0 to 1, not guaranteed to change linearly*/) {
    draw_point(self.path_measure.position_at(progress * self.path_length));
[...]

PR #790, opened 2022-06-20
Add PathMeasure.

Mivik:
Implementation mentioned in #787, with tests and simple documents.

### DIFF
--- a/crates/algorithms/src/lib.rs
+++ b/crates/algorithms/src/lib.rs
@@ -14,6 +14,7 @@ pub mod fit;
 pub mod hatching;
 pub mod hit_test;
 pub mod length;
+pub mod measure;
 pub mod raycast;
 pub mod walk;
 pub mod winding;
--- a/crates/algorithms/src/measure.rs
+++ b/crates/algorithms/src/measure.rs
@@ -1,0 +1,755 @@
+//! Perform cached measurements and split operations on a path.
+//!
+//! # Path measuring
+//!
+//! ## Example
+//!
+//! ```
+//! use lyon_algorithms::math::point;
+//! use lyon_algorithms::path::Path;
+//! use lyon_algorithms::{
+//!     length::approximate_length,
+//!     measure::{MeasureResult, PathMeasure},
+//! };
+//!
+//! fn main() {
+//!     let mut path = Path::builder();
+//!     path.begin(point(0.0, 0.0));
+//!     path.quadratic_bezier_to(point(1.0, 1.0), point(2.0, 0.0));
+//!     path.end(false);
+//!     let path = path.build();
+//!     let mut measure = PathMeasure::from_path(&path, 1e-3);
+//!
+//!     let MeasureResult {
+//!         position, tangent, ..
+//!     } = measure.measure(0.5);
+//!     println!("Mid-point position: {:?}, tangent: {:?}", position, tangent);
+//!
+//!     let last_half = measure.split(0.5..1.0);
+//!     assert!((measure.length() / 2.0 - approximate_length(&last_half, 1e-3)).abs() < 1e-3);
+//! }
+//! ```
+//!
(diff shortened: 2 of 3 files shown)


nical:
That was fast! I'll have look soon.

nical:
Nice work! There's one issue with this PR, though: once the edge has been selected, the `t` value along that edge is approximated by a simple linear interpolation. this works for curves with very small curvatures but the larger the curvature the more imprecise the result is and it can make a pretty big difference.

For example replace the content of `examples/wgpu/src/main.rs` with this https://gist.github.com/nical/3c46ff27abffd2847b6f42af913ae41c and run the example.
The small arrows use the path walker while the large ones use the path measurer. They should be moving at constant speed because the t values are increasing at constant speed, however in the case of the measurer, arrows slow down a lot where the curvature is strongest.

Skia deals with it by not caching an item per path segment, instead they split the curves into smaller sub-curves that are flat enough for the distortion to not be an issue (you can use the tolerance threshold there).
Another solution could be to keep caching entire path segments but flatten the selected segment on the fly when doing the lookup. (tradeoff memory/speed).

Mivik:
Thanks for pointing it out! It was my overlooking the problem. Though having problems running the example on my linux laptop (error details are given as follow), I'll try to address the problem soon.

```
MESA-INTEL: warning: Performance support disabled, consider sysctl dev.i915.perf_stream_paranoid=0

[2022-06-21T15:10:26Z ERROR winit::platform_impl::platform] X11 error: XError {
        description: "BadMatch (invalid parameter attributes)",
        error_code: 8,
        request_code: 149,
        minor_code: 4,
    }
[2022-06-21T15:10:26Z ERROR winit::platform_impl::platform] X11 error: XError {
        description: "BadMatch (invalid parameter attributes)",
        error_code: 8,
        request_code: 149,
        minor_code: 4,
    }
// ...
// [The same message repeating a lot of times]
// ...
thread 'main' panicked at 'Failed to focus input context: XError { description: "BadMatch (invalid parameter attributes)", error_code: 8, request_code: 149, minor_code: 4 }', /home/mivik/.cargo/registry/src/github.com-1ecc6299db9ec823/winit-0.26.1/src/platform_impl/linux/x11/event_processor.rs:884:30
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

And sometimes the error varies as well. In another case, the same message repeats just as above, but the program does not crash immediately. Instead, a black window is shown and `Swap-chain error: Timeout` is repeatedly printed to the console.

This error occurs no matter whether the `examples/wgpu/src/main.rs` is replaced.

Mivik:
Fixed. Now in your example PathMeasure and PathWalker are able to produce the same results.

BTW, in my testing process I noticed a small bug which may cause `PathWalker` to be unable to stop immediately after `Pattern::next` returns `None` so I fixed it.

nical:
Thanks! Looks good to me overall. Please squash this into a single commit, remove test.txt, and I'll merge the PR.
I'll make a few cosmetic changes to the measurer before lyon's next release (nothing big). Also thank you for the walker fix.

Mivik:
Done. So I think I'll close #787 then?

nical:
Woops! A test is failing. I suspect that it is because you now compute the length using only the flattened approximation that is a bit less precise than when approximating the length directly. This shouldn't be a big problem, perhaps the test can be tweaked to use a finer tolerance, the threshold for the assertion be relaxed a bit, or a mix of both.

Mivik:
Strange. All tests have been passed on my laptop. I suspect it has something to do with the floating-point error since I compared them directly in some tests. I'll try to fix it.

Mivik:
OMG. I checked the detail and found that `rustdoc` compiled the example in the document part of `measure.rs`. I never knew that doc-tests even exist and code snippets in document are compiled are executed too, so I just wrote the example without even running it. 😂

nical:
Ah indeed, that one also trips me every now and then. Thanks again!

PR #791, opened 2022-06-24
Fix a bug in StrokeBuilder and a typo in PathMeasure.

Mivik:
In `StrokeBuilder`, `prev` is not set in `begin`. This could lead to unexpected behavior. For example, `StrokeTessellator::tessellate_circle` would call `begin`, `cubic_bezier_to` and so on sequently. If `prev` is not set after `begin` is called, the start point of the cubic bezier would be undefined and it indeed resulted in weird outcomes in my test.

BTW, a typo in my recent pr is fixed.

### DIFF
--- a/crates/algorithms/src/measure.rs
+++ b/crates/algorithms/src/measure.rs
@@ -39,8 +39,8 @@ use crate::path::{
 use std::ops::Range;
 
 #[doc(hidden)]
-pub struct EmptyAttribtueStore;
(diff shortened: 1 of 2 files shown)


nical:
Nice catch! Could you remove the type fix from the patch? I have a work in progress patch locally that does a fair amount of refactoring in the measurer and among other things completely removes EmptyAttributeStore since the unit type `()` already implements AttributeStore in the same way.

Mivik:
No problem.

PR #792, opened 2022-06-24
Fix a bug in StrokeBuilder.

Mivik:
In `StrokeBuilder`, `prev` is not set in `begin`. This could lead to unexpected behavior. For example, `StrokeTessellator::tessellate_circle` would call `begin`, `cubic_bezier_to` and so on sequently. If `prev` is not set after `begin` is called, the start point of the cubic bezier would be undefined and it indeed resulted in weird outcomes in my test.

Since I'm unfamiliar with git so I just closed my last pr to open a new one.


nical:
Thanks a lot!

PR #807, opened 2022-08-05
add `Path::with_attributes` constructor

LU15W1R7H:
It's important to be able to construct an empty `Path` with a certain number of attributes.
I need this in my program.

### DIFF
--- a/crates/path/src/path.rs
+++ b/crates/path/src/path.rs
@@ -113,6 +113,15 @@ impl Path {
         }
     }
 
+    #[inline]
+    pub fn with_attributes(num_attributes: usize) -> Path {
+        Path {
+            points: Box::new([]),
+            verbs: Box::new([]),
+            num_attributes,
+        }
+    }
+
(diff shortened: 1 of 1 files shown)


crates/path/src/path.rs
```diff
@@ -105,11 +105,11 @@ impl Path {
 
     /// Creates an Empty `Path`.
     #[inline]
-    pub fn new() -> Path {
+    pub fn new(num_attributes: usize) -> Path {
```
nical, line 108: Hi, thanks for the PR.

It's a bit early for API breaking changes and since most users of the crate don't use attributes a lot, I'd like the default to remain simple. Could you instead add another constructor `Path::with_attributes(usize) -> Path`?

nical (changes requested):

crates/path/src/path.rs
LU15W1R7H: Done!

nical:
Thanks!

PR #808, opened 2022-08-07
Refactor error system with thiserror crate

YoshieraHuang:
thiserror crate can remove many boilerplate code of error system.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -907,6 +907,7 @@ name = "lyon_extra"
 version = "1.0.1"
 dependencies = [
  "lyon_path",
+ "thiserror",
 ]
 
 [[package]]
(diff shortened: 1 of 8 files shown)


nical:
Hi, thanks for the PR. I'll wait for the next round of breaking changes before merging it.

YoshieraHuang:
> Hi, thanks for the PR. I'll wait for the next round of breaking changes before merging it.

Happy to see this. This PR only rewrite some code in another way and do not change any apis. I may use the word *refactor* in a wrong way.

nical:
Oh sorry you are right, the mistake is mine, I expected that it would introduce breaking changes but indeed it doesn't appeat to.

PR #809, opened 2022-08-07
Resolve cargo clippy warnings

YoshieraHuang:
This PR resolves some clippy warnings and allows another warnings so that there should be no clippy warnings now.

Some clippy warnings which are resolved:
 - literal float precision issue
 - redundant reference
 - redundant slice of the whole range
 - redundant clone
 
Some clippy warnings which are allowed:
 - needless `fn main` in doctest
 - false positive issues

### DIFF
--- a/crates/algorithms/src/walk.rs
+++ b/crates/algorithms/src/walk.rs
@@ -169,10 +169,14 @@ impl<'l> PathWalker<'l> {
         while distance >= self.next_distance {
             if self.num_attributes > 0 {
                 let t2 = t.end * self.next_distance / distance;
-                for i in 0..self.num_attributes {
-                    self.attribute_buffer[i] =
-                        self.prev_attributes[i] * (1.0 - t2) + attributes[i] * t2;
-                }
+                self.attribute_buffer
+                    .iter_mut()
+                    .zip(self.prev_attributes.iter())
+                    .zip(attributes.iter())
+                    .take(self.num_attributes)
+                    .for_each(|((attribute_buffer_i, prev_attribute), attribute)| {
+                        *attribute_buffer_i = prev_attribute * (1.0 - t2) + attribute * t2
+                    });
             }
             x += (self.next_distance - self.leftover) * inv_d;
             let (position, tangent) = pos_cb(x);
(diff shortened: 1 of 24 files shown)


crates/geom/src/quadratic_bezier.rs
```diff
@@ -809,14 +809,14 @@ impl<S: Scalar> QuadraticBezierSegment<S> {
...
+                + self.to.to_vector() * S::value(0.062_612_034))
             .length();
-            let v1 = ((self.to - self.from) * S::value(0.4444444444444444)).length();
-            let v2 = (self.from.to_vector() * S::value(-0.0626120363218102)
-                + self.ctrl.to_vector() * S::value(-0.430331482911935)
-                + self.to.to_vector() * S::value(0.492943519233745))
+            let v1 = ((self.to - self.from) * S::value(0.444_444_45)).length();
+            let v2 = (self.from.to_vector() * S::value(-0.062_612_034)
+                + self.ctrl.to_vector() * S::value(-0.430_331_47)
+                + self.to.to_vector() * S::value(0.492_943_53))
```
nical, line 819: Clippy is technically right here, but I'd like to have S::value take an f64 instead of an f32 at some point (too bad I forgot to do it for 1.0), so I prefere to keep the constants intact until then.

crates/algorithms/src/walk.rs
```diff
@@ -169,10 +169,14 @@ impl<'l> PathWalker<'l> {
...
-                        self.prev_attributes[i] * (1.0 - t2) + attributes[i] * t2;
-                }
+                self.attribute_buffer
+                    .iter_mut()
+                    .zip(self.prev_attributes.iter())
+                    .zip(attributes.iter())
+                    .take(self.num_attributes)
+                    .for_each(|((attribute_buffer_i, prev_attribute), attribute)| {
+                        *attribute_buffer_i = prev_attribute * (1.0 - t2) + attribute * t2
+                    });
```
nical, line 179: I understand the that clippy prefers this formulation but I really don't. Pleas revert this one back to its original loop.

nical (commented):
Lots of nice changes here, thanks! Just a few things before this can be merged.

nical (commented):
Unfortunately changing the input of S::value is a breaking change so I would like to make that change much later and and apply the rest of your PR soon

nyurik:
I made a slightly different set of changes in the #859, which will likely conflict with this one, but it shouldn't change any external API

PR #813, opened 2022-08-11
use drain to clear vector and avoid copy

YoshieraHuang:
In the original version, the contents in `edges_below` are iterated by references, some informations are copied implicitly to create active edge and `edges_below` is cleared soon.

IMO, we can use `edges_below::drain(0..)` iterator to empty this vector and iterate over  the original content with ownership. The result is same and correct. No `clear` is needed and no more copies are performed.

### DIFF
--- a/crates/tessellation/src/fill.rs
+++ b/crates/tessellation/src/fill.rs
@@ -1526,7 +1526,7 @@ impl FillTessellator {
         let from_id = self.current_vertex;
         self.active.edges.splice(
             above,
-            self.edges_below.iter().map(|edge| ActiveEdge {
+            self.edges_below.drain(0..).map(|edge| ActiveEdge {
                 from,
                 to: edge.to,
                 winding: edge.winding,
@@ -1536,8 +1536,6 @@ impl FillTessellator {
                 range_end: edge.range_end,
             }),
         );
-
-        self.edges_below.clear();
     }
 
     fn split_event(&mut self, left_enclosing_edge_idx: ActiveEdgeIdx, left_span_idx: SpanIdx) {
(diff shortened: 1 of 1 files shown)


crates/tessellation/src/fill.rs
```diff
@@ -1526,7 +1526,7 @@ impl FillTessellator {
         let from_id = self.current_vertex;
         self.active.edges.splice(
             above,
-            self.edges_below.iter().map(|edge| ActiveEdge {
+            self.edges_below.drain(0..).map(|edge| ActiveEdge {
```
nical, line 1529: I benchmarked both and the performance is pretty much the same. Note that `edges_below.iter().map` does not copy the edges, they are passed by reference to the callback (while `drain` actually copies them since they are passed by value.

But since there isn't a significant performance difference, might as well take your version which is more self-explanatory.

Just a tiny nit: let's make it `drain(..)` instead of `drain(0..)`.

nical (changes requested):

crates/tessellation/src/fill.rs
YoshieraHuang: The copies I memtioned happen at `map` closure. If the closure take reference, the value must be cloned or copied to new struture. and if the closure take value, the value are moved into new structure. IMO, according to the source code of std, `drain` just set the length of the original vector to right value and take the contents in the given range as a ptr to form a new vector to be iterated over. Ok, copies may perform internally when value are moved into new structure. 

I agree that the copies do not matter in this performance contexts. But a more self-explanatory expression may be welcomed for code viewer.

nical (approved):

PR #827, opened 2022-10-10
Update dead link to Cubic Bezier paper

dnfield:
Unfortunately, the current link redirects to some other page for U. South Alabama.

This link appears to be the same paper.

### DIFF
--- a/crates/geom/src/lib.rs
+++ b/crates/geom/src/lib.rs
@@ -50,7 +50,7 @@
 //! </svg>
 //!
 //! The flattening algorithm implemented in this crate is based on the paper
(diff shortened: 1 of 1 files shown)


nical:
Oops! This crate now uses [Raph Levien's flattening algorithm](https://raphlinus.github.io/graphics/curves/2019/12/23/flatten-quadbez.html) So the link is out of date in more ways than one. Let's merge it so there's a more correct link in the git history, and then I'll remove it.

Thanks!

dnfield:
I did notice that it doesn't seem to actually use Hain's algorithm anymore haha. What made you switch?

PR #830, opened 2022-10-19
Add forbid(unsafe_code) to crates without unsafe code

notgull:
This makes me feel better when this crate appears in my `cargo geiger` graph.

### DIFF
--- a/crates/algorithms/src/lib.rs
+++ b/crates/algorithms/src/lib.rs
@@ -1,6 +1,7 @@
 #![doc(html_logo_url = "https://nical.github.io/lyon-doc/lyon-logo.svg")]
 #![deny(bare_trait_objects)]
 #![allow(clippy::float_cmp)]
+#![forbid(unsafe_code)]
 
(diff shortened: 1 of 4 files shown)


nical:
Hi, sorry for the delay. I've been thinking about this over the past few days, here is how I feel:

If I were to `forbid(unsafe_code)` lyon crates that don't use unsafe, I would be lying because ultimately I make the call whether unsafe code is beneficial or not, and I would simply remove `forbid(unsafe_code)` as soon as I decide that something needs it enough to warrant the risks.

In other words, I think that `forbid(unsafe_code)` should only be used to make it explicit that the project has a policy to never add unsafe code. Otherwise it is pointless because cargo geiger can already tell you that the code does not use unsafe and that's the only thing that ultimately matters for the downstream application.

In lyon, while adding unsafe code should have a strong justification, I am not making any promise that it won't happen. I would prefer to be open and honest about that.

notgull:
I see. I'll close this PR, in that case.

PR #831, opened 2022-10-19
Make several lyon crates no_std

notgull:
This PR makes several crates `no_std`. These crates did not depend on `libstd` for anything save for floating point functions, which are now used through the `num_traits` crate. The `num_traits` crate has its `libm` feature enabled. When the `std` feature is enabled on the affected crates, the `std` feature is in turn enabled on `num_traits`, avoiding the use of `libm` on targets with built-in floating point operations.

The crates I've converted are:

- `lyon_geom`
- `lyon_path` (still uses `alloc`)
- `lyon_algorithms` (still uses `alloc`)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -876,6 +876,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "libm"
+version = "0.2.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "292a948cd991e376cf75541fe5b97a1081d713c618b4f1b9500f8844e49eb565"
+
+[[package]]
 name = "lock_api"
 version = "0.4.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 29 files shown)


nical:
Hi, thanks for this PR. Do you have motivating use case for using lyon crates in a no-std context?

notgull:
I am currently writing a GUI framework that I'd like to be `no_std`, e.g. it's welded directly to libxcb and doesn't need libstd at all (largely for code bloat reasons). I'd like to use these crates to provide geometric primitives for this framework. In order to fit, these crates would need to be no_std.

nical (changes requested):
Sorry for the delay, I've been away from computers for a bit. Thanks for explaining the motivation, sounds good to me.

Just one thing: If I understand correctly, with this change libm will be used by default, I'd rather have libm only used in the no-std mode as libm is reportedly slower for at least some of the provided math utilities.

notgull:
The `num-traits` crate only uses `libm` if `libstd` primitives are not available. Therefore, when the `std` feature is enabled, this should not be an issue.

nical:
You are right, my bad. Good to know.

Issue #846, opened 2022-12-31
Rounded Polygons


wainwrightmark:
It would be nice to have an easy way to create polygons with rounded corners.  
The interface would be similar to that of `Polygon` but would have additional parameters for corner radius and whether the points are specified in clockwise or anticlockwise order.

I've managed to build a rudimentary version of this for my project, I'm happy to expand and contribute it if you think it's a good feature.

nical:
The builders for the default path data structure have a helper for that: https://docs.rs/lyon_path/latest/lyon_path/builder/struct.NoAttributes.html#method.add_rounded_rectangle is it what you are looking for?

wainwrightmark:
I'm looking for the same visual effect as that but for arbitrary polygons (both convex and concave), similar to https://docs.rs/lyon_path/latest/lyon_path/builder/struct.NoAttributes.html#method.add_polygon

nical:
Oh right, sorry, I read your original message a tad too quickly. If your implementation is general enough to work as expected for all polygons, we could add it to lyon_algorithms.
It should probably be implemented as a function that takes a `Polygon` and a `&mut PathBuilder` as parameters.

PR #849, opened 2023-01-04
`add_rounded_polygon`

wainwrightmark:
Added `rounded_polygon::add_rounded_polygon()` to lyon_algorithms.  

Closes https://github.com/nical/lyon/issues/846

I also made the arrows in the wgpu example rounded to show them off but am happy to revert that.


nical (changes requested):
Looks god overall, thanks!

One thing I would like to this to support before merging, is to limit each vertex's radius to half of its smallest edge (no weird thing happening if an edge is less than twice the radius).

PR #852, opened 2023-02-05
Fix the stroke miter-limit (bis)

### LINKED ISSUE #851: Stroke drawing regression in 1.0
Noticed in https://github.com/ruffle-rs/ruffle .

Regression was found in commit https://github.com/nical/lyon/commit/5a452b1cb5312041e3a911725a0757ba5e91769a .

Given SVG shape (original is an SWF, so this SVG is a result of a tool-made dump):
```svg
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<svg xmlns:xlink="http://www.w3.org/1999/xlink" height="334.0px" width="286.2px" xmlns="http://www.w3.org/2000/svg">
  <g transform="matrix(1.0, 0.0, 0.0, 1.0, 20.8, 0.5)">
    <path d="M28.7 284.0 Q33.45 279.2 60.2 269.8 88.0 260.0 99.7 260.0 113.65 260.0 129.7 265.95 L141.3 270.65 Q147.45 273.2 150.55 274.0"
    fill="none" stroke="#000033" stroke-linecap="round" stroke-linejoin="miter-clip" stroke-miterlimit="1.0" stroke-width="14.0"/>
  </g>
</svg>
```

Ruffle draws this:

Expected: 1 commit earlier, I correctly get this:

image

If you want, I can give you the exact builder calls we do, but this will take a bit more time to generate.

nical:
The "fix" I made in #848 was completely wrong.

Fixes #851.

### DIFF
--- a/crates/tessellation/src/stroke.rs
+++ b/crates/tessellation/src/stroke.rs
@@ -1680,7 +1680,9 @@ fn compute_join_side_positions_fixed_width(
     //    flattened segment.
     let extruded_normal = front_normal * vertex.half_width;
     let unclipped_miter = (join.line_join == LineJoin::Miter || join.line_join == LineJoin::MiterClip)
-        && !miter_limit_is_exceeded(extruded_normal, miter_limit);
+        && !miter_limit_is_exceeded(front_normal, miter_limit);
+
+    println!("join {:?} prev tan {prev_tangent:?} next tan {next_tangent:?} normal {normal:?} miter exceeded: {:?}", join.position, miter_limit_is_exceeded(front_normal, miter_limit));
 
     let mut fold = false;
     let angle_is_sharp = next_tangent.dot(prev_tangent) < 0.0;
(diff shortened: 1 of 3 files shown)


crates/tessellation/src/stroke.rs
```diff
@@ -1680,7 +1680,9 @@ fn compute_join_side_positions_fixed_width(
     //    flattened segment.
     let extruded_normal = front_normal * vertex.half_width;
     let unclipped_miter = (join.line_join == LineJoin::Miter || join.line_join == LineJoin::MiterClip)
-        && !miter_limit_is_exceeded(extruded_normal, miter_limit);
+        && !miter_limit_is_exceeded(front_normal, miter_limit);
+
+    println!("join {:?} prev tan {prev_tangent:?} next tan {next_tangent:?} normal {normal:?} miter exceeded: {:?}", join.position, miter_limit_is_exceeded(front_normal, miter_limit));
```
adrian17, line 1685: to be sure, did you mean to leave the println in? :sweat_smile:

nical: oh snap.

nical: Fixed in 1.0.10. Thanks for pointing it out!

PR #857, opened 2023-04-07
Expose path::path::* as path::* for brevity

nyurik:
It used to be that public elements of `lyon::path::*` were exposed, but since 1.0, they were moved to `lyon::path::path::*`.

This PR makes them exposed as `lyon::path::*` again. Eventually you may want to make the path::path private, but that requires a major or at least a minor version bump.

P.S. this makes migrating from pre-1.0 easier too.

### DIFF
--- a/crates/path/src/lib.rs
+++ b/crates/path/src/lib.rs
@@ -56,7 +56,10 @@ pub mod builder;
 pub mod commands;
 mod events;
 pub mod iterator;
+// TODO: remove "pub" on mod path to avoid redundant  "use lyon::path::path::Path" in user code
+//       breaking change would require 1.1 bump?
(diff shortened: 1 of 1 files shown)


nyurik:
@nical what do you think about this PR?

nical:
mostly positive, I need to page back in the reason (if any) why I didn't just export everything initially. I'm a bit busier than usual so I haven't had the time to figure it out yet.

nyurik:
friendly ping :)  (i know you are busy, hope not to disrupt too much :) )

nical:
Sorry about this, it definitely fell off the radar. Let's do this.

nyurik:
thx!

PR #858, opened 2023-04-07
Please consider migrating to 'main' branch

nyurik:
This PR makes it much easier to rename the primary branch to main. Thanks!


nical:
Thanks, I'll look into that eventually. Note for later: the remaining steps are probably described here https://github.com/github/renaming

PR #865, opened 2023-04-18
Update actions on GitHub Actions workflows

striezel:
The following updates are performed:

* update [`actions/checkout`](https://github.com/actions/checkout) to v3
* update [`actions/setup-node`](https://github.com/actions/setup-node) to v3

Still using the older versions of those actions will generate some warning like in this run: https://github.com/nical/lyon/actions/runs/4676216713

> Node.js 12 actions are deprecated. Please update the following actions to use Node.js 16: actions/checkout@v2, actions-rs/toolchain@v1, Swatinem/rust-cache@v1, actions/setup-node@v1. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.

The PR will get rid of those warnings for `actions/checkout` and `actions/setup-node`, because their newer versions use Node.js 16.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -18,7 +18,7 @@ jobs:
           - nightly
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
 
       - uses: actions-rs/toolchain@v1
         with:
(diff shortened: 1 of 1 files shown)


striezel:
I am aware that the wasm job is failing, but it also fails on the master branch without this pull request. See e. g. <https://github.com/nical/lyon/actions/runs/4676216713/jobs/8282243984>. Therefore I assume that this is not an issue with the pull request.

PR #879, opened 2023-10-26
Prevent the tessellator to panic when comparing a sorting key equal to NaN

Dollab:
Motivation : we've noticed a few crashes where the fill tessellator will panic due to unwrapping a None value on the partial comparison of two sort keys.

We'd rather that the library never crashed but instead returned an appropriate Error. In the meantime, returning a default Ordering in such cases could be prevent the crashing of the app. This seems consistent with the way the same problem is handled line 1882.


nical:
Alright! I suspect that dealing with NaNs will be a bit of a whack-a-mole game. Sanitizing them early might be the best way to go in the long run. In the mean time this can't hurt.

PR #887, opened 2023-12-10
Add `no_std` support for `lyon_tessellation`

konall:
- Removed `lyon_tessellation` rigid dependency on default features from `lyon_path`
- Added new crate feature `std`, enabled by default, which enables `std` feature of `lyon_path` dependency
- Replaced imports from `std` with their corresponding imports from `alloc` and `core`
- Required enabling feature `std` to display printed debug assertions
- Annotated some test functions with `#[cfg(test)]` to be explicit about their dependence on `std`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1043,6 +1043,7 @@ dependencies = [
  "float_next_after",
  "lyon_extra",
  "lyon_path",
+ "num-traits",
  "serde",
 ]
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1044,7 +1044,6 @@ dependencies = [
  "lyon_extra",
  "lyon_path",
  "serde",
- "thiserror",
 ]
 
 [[package]]
--- a/crates/tessellation/Cargo.toml
+++ b/crates/tessellation/Cargo.toml
@@ -16,7 +16,7 @@ path = "src/lib.rs"
 
 [features]
 default = ["std"]
-std = ["lyon_path/std"]
+std = ["lyon_path/std", "num-traits/std"]
 serialization = ["serde", "lyon_path/serialization"]
 debugger = []
 profiling = []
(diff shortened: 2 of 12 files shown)


konall:
I've noticed through testing that the fina; remaining blocker on `no_std` support is the dependency on `thiserror`, which currently [does not support it](https://github.com/dtolnay/thiserror/issues/196)

Would you

konall:
Having read a bit more into this, I've successfully gotten `lyon_tessellation` to compile for `no_std` without any breaking changes (as far as I can tell).

The two big changes were:
- hand-writing the error implementations that `thiserror` provided, but feature-gated to `std`- as asserted [here](https://docs.rs/thiserror/latest/thiserror/#details), this is not a breaking change, and I hope that the new `error` module is concise enough that it doesn't introduce any significant maintenance burden
- adding a dependency on `num_traits` for `no_std` floating point and trig operations- as mentioned above, this dependency unfortunately can't be disabled for `std` targets, but I copied the exact same way this is set up in the `lyon_path` crate so I hope it's ok

I'd love to see this merged, so please let me know if there's anything that needs fixing, thanks!

nical:
Hi, thanks for this PR, I haven't had time to look at it yet but I hope to get to it soon.

nical:
Looks good, thanks!

konall:
Thanks! You may have noticed but with this change the entire `lyon` crate is now `no_std` compatible (ignoring the `lyon_extra` crate of course)- it should be trivial to update `lyon` to reflect this but let me know if you want a PR submitted

nical:
I hadn't noticed! I'm planning to add a fair amount of stuff in lyon's meta crate so let's not mark it as no_std compatible now since I don't know that it can remain that way later. For lyon_algorithm, you are welcome to make the change if you want to, but it's also fine to wait and see if anyone needs it (and submits a PR for it).

PR #893, opened 2024-03-08
Upgrade wgpu_svg example to latest version of usvg

hrydgard:
usvg has gone through some rather major API changes, and when trying to integrate code from your svg sample into a prototype I'm working on, I ran into some issues.

This upgrades the svg example to use the very latest usvg, 0.40.

It does not upgrade wgpu and winit and stuff, to keep PR complexity down.

Weirdly, throughout the course of the changes, sometimes the image would be offset by quite a bit, and then it would later come back. Some intricacy about hierachical transforms, I suspect. Oh well, it looks correct with test.svg in the latest commit, at least.


nical:
Thanks!

PR #898, opened 2024-04-20
Better support scientific notation in the path parser

### LINKED ISSUE #895: PathParser: can't parse path using scientific notation for numbers.
`PathParser::parse` fails if the path string contains number using the scientific notation `1e-6`

I don't really know if it's actually valid according to the specs, but browser handle that without problems and I've seen this in the wild (example in https://github.com/slint-ui/slint/issues/5100)

Testcase:

```rust
fn main() {
    let mut parser = lyon_extra::parser::PathParser::new();
    let mut builder = lyon_path::path::Builder::default();
    let source = "M 1.6e-9 1.4e-4 L 10 0 L 10 10 L 0 10 z";
    parser
        .parse(
            &lyon_extra::parser::ParserOptions::DEFAULT,
            &mut lyon_extra::parser::Source::new(source.chars()),
            &mut builder,
        )
        .unwrap();
}
```

Currently fails as parse returns an error `Line 0 Column 5: Expected number, got "e-9"`

nical:
It was able to parse 1e-2 but failed to parse 1.5e-2

Fixes #895.

### DIFF
--- a/crates/extra/src/parser.rs
+++ b/crates/extra/src/parser.rs
@@ -428,6 +428,16 @@ impl PathParser {
             src.advance_one();
         }
 
+        if src.current == '.' {
+            self.float_buffer.push('.');
+            src.advance_one();
+
+            while src.current.is_numeric() {
+                self.float_buffer.push(src.current);
+                src.advance_one();
+            }
+        }
+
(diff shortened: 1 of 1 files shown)


ogoffart:
thanks!

Issue #884, opened 2023-11-27
wGPU example fails on MacOS


tgross35:
It says this is an issue with metal so I don't think it is the same as  #721

```
== wgpu example ==
Controls:
  Arrow keys: scrolling
  PgUp/PgDown: zoom in/out
  b: toggle drawing the background
  a/z: increase/decrease the stroke width
2023-11-27 03:58:23.239 wgpu[80219:4242123] CAMetalLayer ignoring invalid setDrawableSize width=4294967295.000000 height=4294967295.000000
[2023-11-27T08:58:23Z ERROR wgpu::backend::direct] Handling wgpu errors as fatal by default
thread 'main' panicked at /Users/tmgross/.cargo/registry/src/index.crates.io-6f17d22bba15001f/wgpu-0.13.1/src/backend/direct.rs:2391:5:
wgpu error: Validation Error

Caused by:
    In Device::create_texture
      note: label = `Depth texture`
    Dimension X value 4294967295 exceeds the limit of 8192
```

tgross35:
Maybe this would get better if the wgpu dependency was bumped, the example is on 0.13.1 and latest is 0.18.0

PR #902, opened 2024-04-29
Update `wgpu` and `winit` versions in the Wgpu example

tgross35:
These dependencies were quite out of date and were hitting some bugs when running. Update dependencies to the latest and do needed restructuring to make the example run.

Also use the Winit changes as a reason to break the code into a more logical structure, and add a gif to show what the demo does (I had no clue!).

Part of #884

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,3 +52,4 @@
 .vscode
 bin/
 **/target/
+.DS_Store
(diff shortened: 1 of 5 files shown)


tgross35:
This doesn't completely fix #884 since I wasn't able to get to the SVG example. Still that is pretty significant, I am glad to be able to actually run the demo now :)

nical:
Thanks!

PR #904, opened 2024-05-11
Fix PathBuffer::clear not clearing its path list

manpat:
Found this while trying to cache a PathBuffer for reuse.

Without this change, clearing will leave PathBuffer in an invalid state, with `paths` containing stale indices. Any subsequent paths added will either reference incorrect `points` and `verbs` or panic trying to reach out of bounds.

Or in my case panic when the paths I'm building change shape.


nical:
Woops, thanks!

Issue #910, opened 2024-06-12
Line intersection loses lot of accuracy


SkiFire13:
Reproduction: run `cargo run show "M40000 40000 a 12 12 0 0 0 0 -24"` on the main branch (8d6bb2de37765078c66be84f34fc05da483a2e5a)

Result: 
![Screenshot at 24-06-12 ! 09-07-49](https://github.com/nical/lyon/assets/9020423/194f3d1e-0d32-4949-966d-d3a6a1dceee8)

The problem seems to be that the conversion step from arc to quadratic bezier involves computing the control point as the intersection of two lines. However due to the (relatively) high coordinates this results in a non-negligible error that really messes up the curve when you zoom in.

In my usecase I have some svgs where the bottom right corner in this numeric range and the errors are pretty evident in that area.

nical:
Thanks for filing this and providing a simple test case. I don't have a lot of time to spend on this these days so I don't promise a fix soon, but your analysis sounds right and it would be worth avoiding the intersection here.

PR #913, opened 2024-07-12
Use `f64` when approximating SVG arcs with quadratic beziers

SkiFire13:
Helps with #910. This is not a definitive nor general fix, but at least for now avoids the issue when tesselating SVG paths.


nical (approved):
Thanks!

PR #914, opened 2024-07-19
Update wgpu to 22.0

nical:
In the examples and cli app.


nical:
huh I forgot to land this.

Issue #896, opened 2024-04-19
`PathSampler::split_range` seems to disregard discontinuities in the original path


arximboldi:
I am trying to use the `PathSampler` to created dashed patterns as suggested here #673

Here is what my code looks like more or less:
```rust
struct Dash {
    pub gap: Float,
    pub length: Float,
}

fn apply_dashes(
    path: lyon::path::Path,
    dashes: &Vec<Dash>,
    line_width: f32,
) -> lyon::path::Path {
    use lyon_algorithms::{
        measure::{PathMeasurements, SampleType},
        path::Path,
    };

    if dashes.len() == 0 || dashes.len() == 1 && dashes[0].gap == 0.0 {
        return path;
    }

    let measurements = PathMeasurements::from_path(&path, 0.001);
    let mut sampler = measurements.create_sampler(&path, SampleType::Distance);
    let length = sampler.length();
    let mut dashed = Path::builder();
    let mut pos = 0.0;
    while pos < length {
        for dash in dashes {
            sampler.split_range(pos..(pos + dash.length * line_width), &mut dashed);
            dashed.end(false);
            pos += dash.length * line_width + dash.gap * line_width;
        }
    }
    return dashed.build();
}
``` 

When the original path contains discontinuities (e.g. as produced when using Svg builder and doing `move_to()` to move to arbitrary points) the discontinuities do not seem to be respected by `split_range()`, and it create a line connecting the discontinuity.

I am gonna try as a workaround to produce the dashed path in multiple steps (producing one individual path per disjoint segment) but I am concerned with the performance implications. 

Or is there perhaps a mistake in my code or something I am not considering?

nical:
That sounds like a bug in the path sampler. I'll try to have a look soon-ish (no promises, I'm very short on spare time these days).

arximboldi:
Just chiming back in to report that the workaround has worked: now I create a new path for every discontinuous segment and sample that (accumulating the results in a single builder) and this seems to be working. Something around these lines:
```rust
fn apply_dashes<B: lyon::path::builder::PathBuilder>(
    builder: &mut B,
    path: lyon::path::Path,
    dashes: &Vec<Dash>,
    line_width: f32,
) {
    use lyon_algorithms::measure::{PathMeasurements, SampleType};
    let measurements = PathMeasurements::from_path(&path, 0.001);
    let mut sampler = measurements.create_sampler(&path, SampleType::Distance);
    let length = sampler.length();
    let mut pos = 0.0;
[...]

PR #924, opened 2024-12-04
Fix incorrect split_range behaviour on disjointed paths

timmb:
Resolves https://github.com/nical/lyon/issues/896

Previously `split_range` would create incorrect paths in the input path was formed of discontinuous subpaths because neither `add_segment` was not keeping track of whether we're inside a subpath or not. This PR offers the simplest way I could see to resolve.

Co-authored by @nical (for the test).


nical (approved):

nical:
Thanks a lot! Apologies for the delay, let's land this.

Issue #923, opened 2024-11-07
MiterClip flips between Miter and Bevel instead of clipping


timmb:
According to the docs, [`LineJoin`](https://docs.rs/lyon_tessellation/latest/lyon_tessellation/enum.LineJoin.html) should behave the same as in the [SVG specification](https://svgwg.org/specs/strokes/#StrokeLinejoinProperty). This means:

- Miter falls back to Bevel when the miter-limit is exceeded
- MiterClip crops the miter at the point where miter-limit is exceeded.

However, the behaviour I'm seeing with `StrokeTessellator` is that both Miter and MiterClip fall back to Bevel when the limit is exceeded.

## Example

With 

```rs
StrokeOptions {
            start_cap: Butt,
            end_cap: Butt,
            line_join: MiterClip,
            line_width: 30.0,
            variable_line_width: None,
            miter_limit: 7.1,
            tolerance: 0.1,
        }
```

this is the rendered join:
<img width="296" alt="image" src="https://github.com/user-attachments/assets/3e01e43d-31dc-4572-98fe-c01fcfa66eed">

Increasing the miter limit simple changes the point where it flips to the bevel shown.

## Note

This looks the same as https://github.com/nical/lyon/issues/573, however that issue was resolved with a PR, so I'm confused. I'm using `lyon = "1.0.1"`. I also tried specifically using `lyon_tessellation = "1.0.15"`. Perhaps it regressed?

PR #929, opened 2025-04-13
Fix incorrect miter clip length

nickyfahey:
If I'm understanding it correctly `MiterClip` should clip at a distance of the miter limit multiplied by the stroke width, but at the moment it actually clips to a quarter of this because the half width is being halved again instead of doubled in the clip distance calculation.

Possibly related to #923 but I could not reproduce the bug exactly as described (falling back to bevel as opposed to just getting the clip length wrong).

Before fix:
miter-limit-bug-before

After fix:
miter-limit-bug-after

### DIFF
--- a/crates/tessellation/src/stroke.rs
+++ b/crates/tessellation/src/stroke.rs
@@ -3376,3 +3376,40 @@ fn issue_894() {
         &mut BuffersBuilder::new(&mut geometry, VariableWidthStrokeCtor),
     );
 }
+
+#[test]
+fn correct_miter_clip_length() {
+    let path_max_y = 300.0;
+    let line_width = 30.0;
+    let miter_limit = 3.0;
+
+    let mut path = Path::builder();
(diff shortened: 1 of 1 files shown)


nical (changes requested):
Thanks!

According to the [SVG spec](https://svgwg.org/svg2-draft/single-page.html#painting-StrokeMiterlimitProperty):

> the miter is clipped by a line perpendicular to the line bisecting the angle between the two path segments at a distance of half the value of miter length from the intersection of the two path segments. 

So unless I missed something, lyon currently clips at half of the correct distance (oops), and your change makes it clip at twice the correct distance.

The fix should be:

```rust
get_clip_intersections(n0, n1, normal, miter_limit * join.half_width);
```

nickyfahey:
@nical Woops, looks like I was looking at the SVG 1 spec. Fixed it now.

nical (approved):
Thanks!

PR #942, opened 2026-02-22
Fix some vertex index overflow issues in geometry builders

moulins:
- Make sure that using `BuffersBuilder::with_vertex_offset` with large values cannot overflow `OutputVertex` indices.
- Fix `MaxIndex` bounds for `u32, u64, i64, usize, isize`: `VertexId::INVALID` wasn't properly excluded, and `isize` had the entirely wrong value on 32-bit targets.


nical (approved):
Thanks! Sorry for the delay.

PR #956, opened 2026-04-17
Fix zero length when first 2 points are same

fundon:
When the first 2 points are same, `cur.distance - prev.distance` will be **0**.
And it will throw an error `assertion failed: p.x.is_finite()`.

When `0` is detected, it can be returned directly.


nical (approved):
Thanks for this PR and apologies for the delay. Looks good!

Issue #868, opened 2023-05-02
Add Arcs Line Join Type


SonyStone:
https://github.com/nical/lyon/blob/a6e9fc98d25dd5cd6f948edf67a3fa2116e04956/crates/path/src/lib.rs#L189

Is it possible to add another, non-standard stroke-linejoin type?
![image](http://tavmjong.free.fr/SVG/LINEJOIN/svg2_join.png)
![image](https://www.w3.org/TR/2015/WD-SVG2-20150409/images/painting/linejoin-four.svg)

It is called extrapolate or "arcs" join when the line continues along a given path. 

More details at these links:
* https://www.w3.org/TR/2015/WD-SVG2-20150409/painting.html#LineJoin
* http://tavmjong.free.fr/SVG/LINEJOIN/index.html

SonyStone:
With help from AI, I made prototype of a new type of joint for lyon - arcs
<img width="1200" height="1309" alt="Image" src="https://github.com/user-attachments/assets/8e1864db-c87e-4f17-8dca-c26820398272" />

<img width="1208" height="853" alt="Image" src="https://github.com/user-attachments/assets/1902aa82-aa16-4875-a452-5b4013eb3439" />

https://github.com/user-attachments/assets/cc79c73e-988d-44c0-a1f2-a47268f25c0a

SonyStone:
There are some edge cases where the algorithm doesn't know what to do. I'm trying to figure out a way to fix that.

<img width="1180" height="726" alt="Image" src="https://github.com/user-attachments/assets/f2b19516-351d-4e8a-8a1a-b00d5ea7eb0e" />

SonyStone:
I also added a rounded corner to the end of the arcs

<img width="1100" height="800" alt="Image" src="https://github.com/user-attachments/assets/ca01f9a5-ebe4-409a-9410-4d8d099bb7db" />

PR #964, opened 2026-09-04
Add SVG 2 arcs joins with optional rounded miter clipping

SonyStone:
The issue: #868 

## Summary

Adds `LineJoin::Arcs` to `lyon_path` and implements it in `lyon_tessellation`.
Original segment tangents and curvature are retained before flattening. Private
modules construct support geometry and triangulate the join, reusing scratch
allocations.

The construction follows the [SVG 2 Candidate Recommendation, 4 October 2018](https://www.w3.org/TR/2018/CR-SVG2-20181004/painting.html#LineJoinShape):
support-circle intersection and radius adjustment, the excessive-curvature guard,
opposite-parallel rectangle handling, and arc-length miter clipping. Nonnegative
miter limits are accepted, including zero and sub-unit limits.

A second, separable commit adds the nonstandard `ArcsClip::Round` extension.
It fits tangent-continuous biarcs to actual miter-cut directions, without moving
the existing stroke mesh. It is not necessarily a semicircle and may extend past
the miter limit. Invalid or non-convex fits retain the flat cut. SVG behavior
remains the default.

No editor, canvas persistence, support-preview API, historical join modes, or
public profiling switches are included.

## Commit sequence

1. `455ff35` — SVG 2 construction, regression tests, SVG example, and benchmarks.
2. `1692323` — optional rounded clipping, metadata/winding tests, and serialization migration test.

The first commit builds and passes the workspace tests independently. The second
can be considered separately if the extension is outside Lyon's desired scope.

## API

```rust
use lyon_tessellation::{ArcsClip, LineJoin, StrokeOptions};

let svg2 = StrokeOptions::default()
    .with_line_join(LineJoin::Arcs)
    .with_line_width(8.0)
    .with_miter_limit(4.0);

let rounded = svg2.with_arcs_clip(ArcsClip::Round);
```

Compatibility points requiring maintainer review:

- Adding a variant to the existing exhaustive `LineJoin` enum breaks downstream
  exhaustive matches. Release/version policy must be decided before merging.
  The new variant is appended so existing serialized variant indices are not shifted.
- `StrokeOptions` is already non-exhaustive. The second commit adds a field and
  builder method. Missing fields in old JSON default to flat clipping; binary
  serialization compatibility is **not** claimed.
- `MINIMUM_MITER_LIMIT` changes from 1 to 0 for all joins, not only arcs.
- There are no new production dependencies. `serde_json` is a dev dependency for
  the migration test.

## Validation

On Windows x86_64, stable Rust 1.90.0:

- `cargo test --workspace --all-features` passes.
- `cargo check -p lyon_tessellation --no-default-features` passes.
- `cargo check --workspace --target wasm32-unknown-unknown` passes.
- `cargo +nightly test -p lyon_tessellation --all-features` passes.
- Tessellation coverage: 281 unit tests, 9 integration tests, 9 doctests.
- `git diff --check` passes.
- Ordinary Clippy is blocked by the pre-existing `overly_complex_bool_expr`
  diagnostic in `crates/tessellation/src/fill.rs:772`, reproduced on the unchanged
  base. With that lint allowed, the check succeeds; existing test warnings remain.
  No warnings were reported in the newly added arcs modules, integration test, or example.

Linux CI and any maintainer-selected MSRV still need validation. Existing unused
`num_traits::Float` import warnings remain in the no-std build.

Generate a comparison rendered from the actual triangle meshes:

```sh
cargo run -p lyon_tessellation --example arcs -- joins.svg
```

The example compares Round, MiterClip, SVG 2 Arcs, and SVG 2 with rounded cuts,
without relying on browser support for arcs joins.

## Performance: open merge blocker

Do not merge on the assumption that existing joins are unaffected.

The unchanged upstream logo benchmarks were built separately at base
`994526f2766c7e9bc5bb16687232fd7e8999dbdf` and at this candidate. Four complete
alternating pairs were measured with default release settings on an Intel
Core i9-13900H, Windows x86_64, Rust 1.90.0. Medians of reported ns/iteration:

| Existing join | Upstream | Candidate | Change |
| --- | ---: | ---: | ---: |
| Miter | 43,633 | 47,754 | +9.4% |
| Bevel | 46,261.5 | 51,593.5 | +11.5% |
| Round | 56,916 | 66,268 | +16.4% |

This is a local microbenchmark result, not a cross-platform estimate. The
standard-only commit also regressed these cases in an earlier five-pair run.
Profiling and reducing this cost is required before calling the implementation
merge-ready.

Reproduce the existing tests in separate checkouts:

```sh
cargo run -p tess_bench --release -- stroke_
```

The new same-input, reused-buffer benchmarks can be selected with `arcs`,
`round_curved`, and `miter_clip_lines`. Initial smoke measurements were noisy;
they are not sufficient to quantify the overhead of rounded clipping.

## Remaining correctness/scope questions

- The SVG comparison example currently combines triangles into one nonzero-filled
  path without normalizing their winding. Overlapping triangles with opposite
  winding cancel, producing apparent holes. Coordinate checks confirm coverage
  in those regions. The exporter needs correction, and mixed mesh winding needs
  a separate review for back-face culling. This has not been fixed in these commits.
- Degenerate or numerically unmeshable joins currently use a round safety
  fallback. SVG 2 describes a line construction for degenerate Bezier handles;
  that behavior still needs reconciliation and dedicated conformance tests.
  This PR does not claim complete SVG conformance.
- Variable-width supports use width at the join, not derivatives of the width
  profile.
- Pre-flattened paths have no curvature and use the straight-segment behavior.
  Elliptical arcs converted to Beziers use those Beziers' curvature.
- Existing stroke self-overlap/alpha-compositing limitations remain.

Please review the public API and desired scope of the rounded-cut extension while
the performance and degenerate-handle questions are addressed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1014,6 +1014,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "itoa"
+version = "1.0.18"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8f42a60cbdf9a97f5d2305f08a87dc4e09308d1276d28c869c684d7777685682"
+
+[[package]]
 name = "jni"
 version = "0.21.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1221,6 +1227,7 @@ dependencies = [
  "lyon_path",
  "num-traits",
  "serde",
+ "serde_json",
 ]
 
 [[package]]
@@ -2006,6 +2013,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "ryu"
+version = "1.0.23"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9774ba4a74de5f7b1c1451ed6cd5285a32eddb5cccb8cc655a4e50009e06477f"
+
+[[package]]
 name = "same-file"
 version = "1.0.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2057,6 +2070,18 @@ dependencies = [
  "proc-macro2",
  "quote",
  "syn",
+]
+
+[[package]]
+name = "serde_json"
+version = "1.0.143"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d401abef1d108fbd9cbaebc3e46611f4b1021f714a0597a71f41ee463f5f4a5a"
+dependencies = [
+ "itoa",
+ "memchr",
+ "ryu",
+ "serde",
 ]
 
 [[package]]
--- a/bench/tess/src/arcs.rs
+++ b/bench/tess/src/arcs.rs
@@ -1,0 +1,79 @@
+//! Same-input comparisons with reused tessellator and output allocations.
+
+use bencher::Bencher;
+use lyon::math::{point, Point};
+use lyon::path::Path;
+use lyon::tessellation::geometry_builder::{simple_builder, VertexBuffers};
+use lyon::tessellation::{LineJoin, StrokeOptions, StrokeTessellator};
+
+fn arcs_curved(bench: &mut Bencher) {
+    run(bench, false, LineJoin::Arcs, 4.0);
+}
+
+fn arcs_clipped(bench: &mut Bencher) {
+    run(bench, false, LineJoin::Arcs, 1.5);
+}
+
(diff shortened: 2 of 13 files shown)


SonyStone:
Replaced by #965 after renaming the source branch to `feature/868-add-arcs-line-join-type`. Both implementation commits are unchanged.

PR #968, opened 2026-09-19
fix compilation failure without default-features

White-Green:
When removing Cargo.lock and performing dependency resolution, specifying `--no-default-features` causes the `lyon_geom` crate to fail to compile.

```
$ rm Cargo.lock
$ cargo check --no-default-features -p lyon_geom
warning: bench/path/Cargo.toml: `package.edition` is unspecified, defaulting to `2015` while the latest is `2024`
warning: `path_bench` (manifest) generated 1 warning
warning: bench/geom/Cargo.toml: `package.edition` is unspecified, defaulting to `2015` while the latest is `2024`
warning: `geom_bench` (manifest) generated 1 warning
warning: bench/tess/Cargo.toml: `package.edition` is unspecified, defaulting to `2015` while the latest is `2024`
warning: `tess_bench` (manifest) generated 1 warning
    Updating crates.io index
     Locking 349 packages to latest compatible versions
      Adding clap v2.34.0 (available: v4.6.7)
      Adding clap v3.2.25 (available: v4.6.7)
      Adding env_logger v0.7.1 (available: v0.11.11)
      Adding float_next_after v1.0.0 (available: v2.0.0)
      Adding itertools v0.7.11 (available: v0.15.0)
      Adding rand v0.6.5 (available: v0.10.2)
      Adding tiny-skia-path v0.11.4 (available: v0.12.0)
      Adding usvg v0.40.0 (available: v0.48.1)
      Adding wgpu v25.0.2 (available: v30.0.1)
   Compiling libm v0.2.16
   Compiling autocfg v1.5.1
    Checking arrayvec v0.7.8
   Compiling num-traits v0.2.19
    Checking euclid v0.22.14
    Checking lyon_geom v1.0.19 (/home/whitegreen/projects/lyon/crates/geom)
error[E0432]: unresolved import `euclid::Trig`
   --> crates/geom/src/lib.rs:114:20
    |
114 |     pub(crate) use euclid::Trig;
    |                    ^^^^^^^^----
    |                            |
    |                            no `Trig` in the root
    |
note: found an item that was configured out
   --> /home/whitegreen/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/euclid-0.22.14/src/lib.rs:63:22
    |
 62 | #[cfg(any(feature = "std", feature = "libm"))]
    |          ----------------------------------- the item is gated here
 63 | pub use crate::trig::Trig;
    |                      ^^^^

For more information about this error, try `rustc --explain E0432`.
error: could not compile `lyon_geom` (lib) due to 1 previous error
```

This PR fixes the issue by enabling the `libm` feature.


nical:
Hi, thanks for this PR, unfortunately libm, while good as a fallback for portability, is slower than builtin trigonometric functions so enabling it by default for all users is not the way to go.

It would be nice if we could enable euclid's libm feature only when in geom is build in a no-std/no-default-features mode.

nical:
The workaround is for the project that needs lyon_geom without std to add a dependency to euclid with the libm feature, which is a bit annoying. Would adding a libm feature to lyon_geom serve your purpose?

White-Green:
Thank you for your response.

Within the euclid crate, the `#[cfg(any(feature = "std", feature = "libm"))]` directive is used to reference features, while the actual libm implementation is handled by the num-traits crate.
https://github.com/servo/euclid/tree/7bd06640b0391d0639fa9f45839be2ef22f79355

The implementation in num-traits appears to use `#[cfg(feature = "std")]` for std implementations and `#[cfg(all(not(feature = "std"), feature = "libm"))]` for libm implementations.
https://github.com/rust-num/num-traits/blob/7ec3d41d39b28190ec1d42db38021107b3951f3a/src/float.rs#L2077-L2172
This already establishes a scenario where libm is used only when std is not available. From a performance perspective, I believe this should be acceptable.

While we must accept the potential issues of additional linking symbols and supply chain attack vulnerabilities introduced by including libm as a dependency, I believe this represents a reasonable compromise.

nical:
I stand corrected, thanks!

